PR #5, opened 2020-05-06
All aliases should be parents of relevant secret key

snarlysodboxer:
This one makes sure all aliases are dependency tree parents on the secret key they belong to, so that when we go to generate the keystore itself, the aliases are all ready.

Also fixed a typo and replaced placeholders that were causing tests to fail.

### DIFF
--- a/pkg/memorystore/memorystore.go
+++ b/pkg/memorystore/memorystore.go
@@ -80,8 +80,25 @@ func addParentsAndChildren(parent, path []string, secretConfig *types.SecretConf
 					if Equal(node.Path, path) {
 						node.Parents = append(node.Parents, parentNode)
 						parentNode.Children = append(parentNode.Children, node)
+						break
 					}
 				}
+			}
+		}
+	}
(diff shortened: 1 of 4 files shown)


pkg/memorystore/memorystore.go
```diff
@@ -80,11 +80,28 @@ func addParentsAndChildren(parent, path []string, secretConfig *types.SecretConf
...
 						node.Parents = append(node.Parents, parentNode)
 						parentNode.Children = append(parentNode.Children, node)
+						break
 					}
 				}
 			}
 		}
 	}
+	// all aliases should be parents of the relevant secret key
+	for _, node := range nodes {
```
maxres-fr, line 90: does this path get tested at all?

snarlysodboxer: Good question, and yes it does. That's what's in the pkg/memorystore/test/helpers.go changes. I test drove the changes, first updating the tests, then updating the code to make the tests pass.

PR #8, opened 2020-05-08
Add EnsureAcyclic

snarlysodboxer:
EnsureAcyclic ensures the config doesn't have circular dependencies, and informs the user if so.

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -70,7 +70,7 @@ type AliasConfig struct {
 	CommonName     string   `yaml:"commonName" validate:"required"`
 	Sans           []string `yaml:"sans"`
 	SignedWithPath []string `yaml:"signedWithPath"`
-	PasswordPath   []string `yaml:"passwordPath" validate:"required"`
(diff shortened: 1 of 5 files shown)


pkg/types/types.go
```diff
@@ -70,7 +70,7 @@ type AliasConfig struct {
 	CommonName     string   `yaml:"commonName" validate:"required"`
 	Sans           []string `yaml:"sans"`
 	SignedWithPath []string `yaml:"signedWithPath"`
-	PasswordPath   []string `yaml:"passwordPath" validate:"required"`
+	PasswordPath   []string `yaml:"passwordPath"`
```
snarlysodboxer, line 73: This is a fix for a bug from a previous PR. This could've been caught earlier by tests against main.go.

secretsConfig.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-config:
+appConfig:
```
snarlysodboxer, line 2: This is a fix for a bug from previous changes. Maybe tests against main.go could include using the provided example config.

PR #11, opened 2020-05-13
Create operator and CRD using kubebuilder

jrcast:
This PR changes the app to work as a K8s operator. I used kubebuilder to generate the boiler plate code. 
I restructured the module's layout. Most of the main.go code is now under the ./controller and the types.go are part of the api/v1alpha1


api/v1alpha1/api_validation.go
```diff
@@ -0,0 +1,148 @@
...
+					}
+				}
+			}
+		}
+	}
+}
+
+func pathExistsInSecretConfigs(path []string, secrets []*SecretConfig) bool {
+	found := false
+path:
```
maxres-fr, line 127: has `go fmt` been run here?

jrcast: is part of the make

controllers/secretagentconfiguration_controller.go
```diff
@@ -0,0 +1,230 @@
...
+	for index := range config.Secrets {
+		config.Secrets[index].Namespace = instance.Namespace
+	}
+
+	//TODO Change to validating webhook: https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation.html
+	validate := validator.New()
+	validate.RegisterStructValidation(secretagentv1alpha1.ConfigurationStructLevelValidator, secretagentv1alpha1.Configuration{})
+	err = validate.Struct(config)
+	if err != nil {
+		log.Error(err, "error validating configuration file: %+v")
```
snarlysodboxer, line 87: Shouldn't this be a `return ctrl.Result{}, err`?
Same with lines 92, 98, 107, and so on.

config/manager/manager.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: system
```
snarlysodboxer, line 6: `name` should always be first under `metadata` IMO for readability/scan-ability. Silly generation code.

config/crd/bases/secret-agent.secrets.forgerock.io_secretagentconfigurations.yaml
```diff
@@ -0,0 +1,211 @@
...
+                            description: AliasConfig is the configuration for a keystore
+                              alias
+                            properties:
+                              algorithm:
+                                description: Algorithm Specifies which keystore algorithm
+                                  to use Only one of the following algorithms may
+                                  be specified.
+                                enum:
+                                - ECDSAWithSHA256
+                                - SHA256withRSA
```
snarlysodboxer, line 100: This is the same as oneof, nice work.

api/v1alpha1/api_validation.go
snarlysodboxer: This is a go label.

snarlysodboxer: That said, ideally your editor should be setup to go fmt and go vet on every file save. It makes development much easier by auto-indenting code when you move it around, and by pointing out mistakes early on.

Makefile
```diff
@@ -0,0 +1,84 @@
+
```
snarlysodboxer, line 1: Is this makefile part of the boiler plate generation? It's good stuff, but I'm worried it also teaches the developer to not have to actually understand what's going on when working with this app.

Makefile
```diff
@@ -0,0 +1,84 @@
...
+manager: generate fmt vet
+	go build -o bin/manager main.go
+
+# Run against the configured Kubernetes cluster in ~/.kube/config
+run: generate fmt vet manifests
+	go run ./main.go
+
+# Install CRDs into a cluster
+install: manifests
+	kustomize build config/crd | kubectl apply -f -
```
snarlysodboxer, line 30: This kind of thing concerns me. What if the user assumes that `make install` installs the go binary locally, and ends up modifying their cluster instead.

snarlysodboxer: Maybe this kind of thing would be better in the readme.

Makefile
```diff
@@ -0,0 +1,84 @@
...
+
+# Uninstall CRDs from a cluster
+uninstall: manifests
+	kustomize build config/crd | kubectl delete -f -
+
+# Deploy controller in the configured Kubernetes cluster in ~/.kube/config
+deploy: manifests
+	cd config/manager && kustomize edit set image controller=${IMG}
+	kustomize build config/default | kubectl apply -f -
```
snarlysodboxer, line 39: For all the reading through this file to figure out what it does and doesn't do in `deploy` vs `install` vs other commands, maybe they should just read the readme? This way the onus is on them if they mess up the commands. And this way they learn.

Makefile
```diff
@@ -0,0 +1,84 @@
...
+vet:
+	go vet ./...
+
+# Generate code
+generate: controller-gen
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+
+# Build the docker image
+docker-build: test
+	docker build . -t ${IMG}
```
snarlysodboxer, line 63: Reverse engineering this file to figure out how to set my custom image tag seems harder than just running the overly familiar `docker build -t my-repo/my-image:my-tag` command.

Makefile
```diff
@@ -0,0 +1,84 @@
...
+controller-gen:
+ifeq (, $(shell which controller-gen))
+	@{ \
+	set -e ;\
+	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\
+	cd $$CONTROLLER_GEN_TMP_DIR ;\
+	go mod init tmp ;\
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.5 ;\
+	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
+	}
```
snarlysodboxer, line 80: What is the reason the user can't just run `go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.5` here if they haven't previously installed it?

Dockerfile
```diff
@@ -18,11 +41,13 @@ RUN addgroup --gid 11111 forgerock && \
...
+COPY --from=builder --chown=forgerock:root /workspace/manager /
 
 RUN mkdir -p /opt/gen/secrets/generic/truststore && \
     cp $JAVA_HOME/lib/security/cacerts /opt/gen/secrets/generic/truststore && \
     chmod 764 /opt/gen/secrets/generic/truststore/cacerts && \
     chown -R forgerock:root /opt/gen
 
 USER forgerock
+ENTRYPOINT ["/manager"]
+
```
snarlysodboxer, line 53: Not everyone agrees with me on this, but over the 6+ years I've been working with Docker, I started setting ENTRYPOINT to bash in almost all my Dockerfiles. The reason is, in every instance where I am starting up the image manually, it's to debug something, so I _always_ end up adding `--entrypoint bash` to my `docker run` command. - And in every place where the image is started automatically for real usage, (whether it's a `docker-compose.yaml`, `deployment.yaml`, or something else,) I want to see the entrypoint explicitly set there so I can see the entire definition of what's being run all in one place. - E.G. env vars, entrypoint, command, volume mounts, - everything.

Having to go look in a Dockerfile when you're looking a `deployment.yaml` to find out what's actually being run is a drag. Having to override the entrypoint whenever you run the image manually is also a drag.

Dockerfile
```diff
@@ -1,13 +1,36 @@
...
+# and so that source changes don't invalidate our downloaded layer
 RUN go mod download
-COPY . /go/src/github.com/ForgeRock/secret-agent/
-RUN go build
+
+# Copy the go source
+COPY main.go main.go
+COPY api/ api/
+COPY controllers/ controllers/
+COPY pkg/ pkg/
```
snarlysodboxer, line 19: Is there a reason here not to just do `COPY . .`? The saved steps saves build time. Also, there's `.dockerignore` if there's things you want to keep out.

.gitignore
```diff
@@ -1,15 +1,26 @@
...
 # Binaries for programs and plugins
 *.exe
 *.exe~
 *.dll
 *.so
 *.dylib
+bin
 
-# Test binary, built with `go test -c`
+# Test binary, build with `go test -c`
```
snarlysodboxer, line 10: This excludes binaries that are built with `go test -c`. It does not build them.

.gitignore
```diff
@@ -1,15 +1,26 @@
...
+
+!vendor/**/zz_generated.*
+
+# editor and IDE paraphernalia
+.idea
+*.swp
+*.swo
+*~
+/.project
+/.vscode/
```
snarlysodboxer, line 26: Good stuff for a gobal gitignore, but it doesn't hurt having them in here as well.

api/v1alpha1/api_validation.go
maxres-fr: yea the indentation just looked off to me.

controllers/secretagentconfiguration_controller.go
```diff
@@ -0,0 +1,230 @@
...
+		// on deleted requests.
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	//Old main.go
+
+	// TODO Datatype workaround
+	data, err := yaml.Marshal(instance.Spec)
+	config := &secretagenttypes.Configuration{}
+	err = yaml.Unmarshal(data, config)
```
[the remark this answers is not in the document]
jrcast: Yep. That's one of the 2 workarounds. This will be fixed in my next push

Makefile
jrcast: The makefile is autogenerated. It has really useful cases but they are poorly named. That's something we can easily change. Top ones to know:
`make manifests`: creates the CRD
`make install`: install the CRD in your kubeconfig cluster
`make docker-build IMG=controller:V0.0.1` Builds the docker image with specified tag
`make docker-push IMG=controller:V0.0.1` push specified image
`make deploy IMG=controller:V0.0.1` deploys the controller with specified image tag
`make` builds the go app
`make run` runs your go app using kubeconfig creds

Makefile
jrcast: Yes boiler plate generation. They have really detailed documentation here: https://book.kubebuilder.io/. It explains quite well all the generated components and how they connect with each other. 
Note: I will be cleaning up some unused files in the  ./config  folder. Mostly kustomize files

Makefile
jrcast: We can certainly revisit this. Do note that the typical user will not run the make file to deploy to their cluster.  I see this make file for dev purposes only. I think the user will just use the Kustomize files directly (we can bundle these files and attach them to the github release 😉 ) or do something like this:


$ kubectl apply --validate=false -f https://github.com/jetstack/cert-manager/releases/download/v0.15.0/cert-manager.yaml

//IMHO I think that's the way to go. We just put that in our README

snarlysodboxer: Sounds good. I can't imagine why we'd want to skip validation though, seems like it's asking for trouble?

snarlysodboxer: BTW, I was meaning the developers when I said user.

jrcast: yeah. I just copy/pasted the command from cert-manager.io. Not sure why they have it that way. 🤷‍♂️

PR #12, opened 2020-05-14
WIP: Keytool integration

snarlysodboxer:
WIP


pkg/generator/keytool.go
```diff
@@ -0,0 +1,193 @@
...
+
+const (
+	// TODO make these configurable with flags
+	dsKeyMgrPath = "/opt/gen/opendj/bin/dskeymgr"
+	keytoolPath  = "/usr/local/openjdk-11/bin/keytool"
+)
+
+var (
+	tempDir          = ""
+	keystoreFilePath = ""
```
maxres-fr, line 22: why using init() for this?

jrcast:
I think we need to get a hold of Simon for this change? I think we need better names for the ca types 😕  WDYT?

pkg/generator/keytool.go
snarlysodboxer: valid question. i think i'm doing this because `ioutil.TempDir` creates a tempdir with a random name when the application starts up, and `keystoreFilePath` needs to contain that name. is there a better way i can do this?

snarlysodboxer: (i'm going to submit another PR soon and then ditch this PR, but that PR will have the same change.)

snarlysodboxer:
Superseded by https://github.com/ForgeRock/secret-agent/pull/25

PR #14, opened 2020-05-19
Fix double base64 encoding.

### LINKED ISSUE #13: Keys in secret are getting double base64 encoded
Take a look at `username`: `WVcxaFpHMXBiZz09` -> `YW1hZG1pbg==` -> `amadmin`

```
apiVersion: v1
Metadata:
  name: am-runtime
  namespace: default
data:
  keypass: WkhodWNGSjVaR3BaU0hwRVQyOHdUQT09
  username: WVcxaFpHMXBiZz09
kind: Secret

```

Steps to reproduce: 
1. start minikube
1.  `make install`
1. `make run`
1. `kubectl create config/samples`
1. see generated secrets

jrcast:
1. Fixes #13 : Secret data was getting double base64 encoded
1. Added info logs when creating/updating secrets.

### DIFF
--- a/controllers/secretagentconfiguration_controller.go
+++ b/controllers/secretagentconfiguration_controller.go
@@ -121,29 +121,25 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
 	}
 
 	//end old main.go
-
-	k8sSecretList := k8ssecrets.GenerateSecretAPIObjects(instance.Spec.Secrets)
-
 	labelsForSecretAgent := func(name string) map[string]string {
 		return map[string]string{"managed-by-secret-agent": "true", "secret-agent-configuration-name": name}
 	}
 
 	if instance.Spec.AppConfig.CreateKubernetesObjects {
-		var operation *string
-		var err error
-		for _, secret := range k8sSecretList {
-			// Set SecretAgentConfiguration instance as the owner and controller of the secret
-			if err := ctrl.SetControllerReference(&instance, secret, reconciler.Scheme); err != nil {
-				return ctrl.Result{}, err
-			}
-			secret.Labels = labelsForSecretAgent(instance.Name)
+		for _, secret := range instance.Spec.Secrets {
+			k8sSecret := k8ssecrets.GenerateSecretAPIObjects(secret)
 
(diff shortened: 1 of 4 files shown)


controllers/secretagentconfiguration_controller.go
```diff
@@ -120,22 +121,29 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
...
-
 	labelsForSecretAgent := func(name string) map[string]string {
 		return map[string]string{"managed-by-secret-agent": "true", "secret-agent-configuration-name": name}
 	}
 
 	if instance.Spec.AppConfig.CreateKubernetesObjects {
-		for _, secret := range k8sSecretList {
-			// Set SecretAgentConfiguration instance as the owner and controller of the secret
-			if err := ctrl.SetControllerReference(&instance, secret, reconciler.Scheme); err != nil {
+		var operation *string
```
maxres-fr, line 129: do we really need a pointer to a string here?

pkg/k8ssecrets/k8ssecrets_test.go
```diff
@@ -110,20 +110,22 @@ func TestApplySecrets(t *testing.T) {
...
 	scheme := runtime.NewScheme()
 	clientgoscheme.AddToScheme(scheme)
-	client := fake.NewFakeClientWithScheme(scheme)
-	err := ApplySecrets(client, k8sSecrets)
-	if err != nil {
+	client := fake.NewFakeClientWithScheme(scheme, k8sSecret2)
+	//k8sSecret1 should be created, k8sSecret2 should be updated
+	k8sSecret2.Data = map[string][]byte{"otherkey": []byte(`YWRtaW4=`)}
+	k8sSecrets := []*corev1.Secret{k8sSecret1, k8sSecret2}
+	if _, err := ApplySecrets(client, k8sSecret1); err != nil {
```
maxres-fr, line 123: if we have to add many more sections like this we should use a "test table".

controllers/secretagentconfiguration_controller.go
jrcast: I think is nicer to compare against nil. i.e. `if operation != nil {`
rather than `if len(operation) > 0 {` if I were to return a string 🤷

jrcast: changed to string

PR #17, opened 2020-05-22
feat: load existing secrets from Azure key vault

maxres-fr:


pkg/secretsmanager/secretsmanager_integration_test.go
```diff
@@ -91,16 +94,21 @@ func TestLoadGCPSecrets(t *testing.T) {
...
 		t.Fatalf("Expected 'admin', got: %s", string(node.Value))
 	}
+
 }
 
 func TestEnsureGCPSecrets(t *testing.T) {
+	gcpProjectID, ok := os.LookupEnv("GOOGLE_CLOUD_PROJECT")
+	if !ok {
+		t.Fatalf("GOOGLE_CLOUD_PROJECT environment variable required")
+	}
```
snarlysodboxer, line 111: We could move this to an init function to avoid the repetition.

pkg/secretsmanager/secretsmanager_integration_test.go
```diff
@@ -148,3 +156,76 @@ func TestEnsureGCPSecrets(t *testing.T) {
...
+		Namespace: "default",
+		Keys:      []*types.KeyConfig{keyConfig},
+	}
+	node := &types.Node{
+		Path:         []string{"load-azure-secrets", "username"},
+		SecretConfig: secretConfig,
+	}
+	nodes := []*types.Node{node}
+
+	vaultName := "secret-agent-test"
```
snarlysodboxer, line 188: Should this also be configurable via env var?

pkg/secretsmanager/secretsmanager_integration_test.go
```diff
@@ -148,3 +156,76 @@ func TestEnsureGCPSecrets(t *testing.T) {
...
+	// expect nothing found
+	baseCtx := context.Background()
+	ctx1, cancel := context.WithTimeout(baseCtx, 3000000000)
+	defer cancel()
+	client, err := newCliAzureClient()
```
snarlysodboxer, line 193: If we're ignoring the error here on purpose, we should assign it to `_`, otherwise check it.

pkg/secretsmanager/secretsmanager_integration_test.go
```diff
@@ -148,3 +156,76 @@ func TestEnsureGCPSecrets(t *testing.T) {
...
+	err = loadAzureSecrets(ctx1, client, vaultName, nodes)
+	if err != nil {
+		t.Fatalf("Expected no error, got: %+v", err)
+	}
+	if len(node.Value) != 0 {
+		t.Fatalf("Expected no value, got: %+v", node.Value)
+	}
+
+	// create a secret to load
+	var secParams keyvault.SecretSetParameters
```
snarlysodboxer, line 203: If you want to avoid the mixed variable instantiation syntax here, you can `secParams := &keyvault.SecretSetParameters{}`.

snarlysodboxer: Personally I like to use the `myvar := blah` syntax inside functions, and the `var myvar blah` syntax outside functions (since it's required outside functions.)

pkg/secretsmanager/secretsmanager_integration_test.go
```diff
@@ -148,3 +156,76 @@ func TestEnsureGCPSecrets(t *testing.T) {
...
+	secName := getSecretID(node.SecretConfig.Namespace, node.Path)
+	secValue := "test-value"
+	secParams.Value = &secValue
```
snarlysodboxer, line 206: Maybe it's because I haven't had enough coffee yet, or because often glance over multiple lines at once instead of always reading linearly, but at first glance I read `secParams` as "security params". In my mind full var names would be easier on the eye to read, like `secretName`, `secretValue`, and `secretparams`. But that's just me. :)

pkg/secretsmanager/secretsmanager_integration_test.go
```diff
@@ -148,3 +156,76 @@ func TestEnsureGCPSecrets(t *testing.T) {
...
+		}
+	}()
+
+	// test loaded secrets
+	err = loadAzureSecrets(ctx1, client, vaultName, nodes)
+	if err != nil {
+		t.Fatalf("Expected no error, got: %+v", err)
+	}
+	if string(node.Value) != secValue {
+		t.Fatalf("Expected value to be: %s but found: %+v", secValue, node.Value)
```
snarlysodboxer, line 228: I think I would use `"Expected value to be: %s but found: %s", secValue, string(node.Value)` here to make the comparison easier.

pkg/secretsmanager/secretsmanager_integration_test.go
```diff
@@ -148,3 +156,76 @@ func TestEnsureGCPSecrets(t *testing.T) {
...
+
+	// create a secret to load
+	var secParams keyvault.SecretSetParameters
+	secName := getSecretID(node.SecretConfig.Namespace, node.Path)
+	secValue := "test-value"
+	secParams.Value = &secValue
+	vaultURL := fmt.Sprintf(azureVaultURLFmt, vaultName)
+	_, err = client.SetSecret(baseCtx, vaultURL, secName, secParams)
+	if err != nil {
+		fmt.Printf("unable to add/update secret: %v\n", err)
```
snarlysodboxer, line 210: You get more information about the error with `%+v`. I pretty much exclusively use that for errors.

snarlysodboxer:
Oh, we should also update the CRD to include allowance for the Azure secret manager settings.

pkg/secretsmanager/secretsmanager.go
```diff
@@ -53,6 +65,46 @@ func EnsureSecrets(ctx context.Context, config *types.Configuration, nodes []*ty
...
+	return nil
+}
+
+// loadAzureSecret load a secret from Azure Key Vault
+func loadAzureSecret(ctx context.Context, client *keyvault.BaseClient, vaultName string, node *types.Node) error {
+	secretID := getSecretID(node.SecretConfig.Namespace, node.Path)
+	response, err := client.GetSecret(ctx, fmt.Sprintf(azureVaultURLFmt, vaultName), secretID, "")
+	if err != nil {
+		if e, ok := err.(autorest.DetailedError); ok && e.StatusCode.(int) == 404 {
+			return nil
```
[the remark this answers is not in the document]
maxres-fr: @snarlysodboxer I'm a bit new to go but `int(e.StatusCode)`  is an interface, I'm not sure I know another way?
```
pkg/secretsmanager/secretsmanager.go:96:54: cannot convert e.StatusCode (type interface {}) to type int: need type assertion
```

pkg/secretsmanager/secretsmanager.go
```diff
@@ -53,6 +65,46 @@ func EnsureSecrets(ctx context.Context, config *types.Configuration, nodes []*ty
...
+	client := keyvault.New()
+	client.Authorizer = authorizer
+	return &client, nil
+}
+
+// loadAzureSecrets loads any existing secrets in Azure Key Valut into the memory store
+func loadAzureSecrets(ctx context.Context, client *keyvault.BaseClient, vaultName string, nodes []*types.Node) error {
+	for _, node := range nodes {
+		err := loadAzureSecret(ctx, client, vaultName, node)
+		return err
```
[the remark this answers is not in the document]
maxres-fr: :facepalm:

pkg/secretsmanager/secretsmanager_integration_test.go
maxres-fr: `AZURE_VAULT_NAME`?

pkg/secretsmanager/secretsmanager_integration_test.go
maxres-fr: me too; copy pasta...

pkg/secretsmanager/secretsmanager_integration_test.go
maxres-fr: yea that's not a good convention, I've updated.

pkg/secretsmanager/secretsmanager.go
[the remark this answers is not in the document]
snarlysodboxer: Super weird that they make `StatusCode` an interface. Not something I see Google or Kubernetes or other major Go projects doing. Anyway, you're right, you did the right thing with the strange way they wrote the code. It would however panic if for some reason `StatusCode`s value was not an int.

pkg/secretsmanager/secretsmanager_integration_test.go
snarlysodboxer: 👍

api/v1alpha1/secretagentconfiguration_types.go
```diff
@@ -112,9 +112,10 @@ const (
 
 // SecretsManager Strings
 const (
-	SecretsManagerNone SecretsManager = "none"
-	SecretsManagerGCP  SecretsManager = "GCP"
-	SecretsManagerAWS  SecretsManager = "AWS"
+	SecretsManagerNone  SecretsManager = "none"
+	SecretsManagerGCP   SecretsManager = "GCP"
+	SecretsManagerAWS   SecretsManager = "AWS"
+	SecretsManagerAzure                = "Azure"
```
snarlysodboxer, line 118:
```suggestion
	SecretsManagerAzure  SecretsManager  = "Azure"
```
This should be of type `SecretsManager` in order to pass validation.

config/crd/bases/secret-agent.secrets.forgerock.io_secretagentconfigurations.yaml
```diff
@@ -69,6 +71,7 @@ spec:
                   - AWS
                   type: string
               required:
+              - azureVaultName
```
snarlysodboxer, line 74: This doesn't seem like it should be required. E.G. if you're using GCP Secret Manager, you shouldn't need to set Azure values. Instead, we can use the custom validation logic to make sure it's set if you selected Azure and the SM.

BTW, this file is not supposed to be modified directly, but instead via the _types.go file.

pkg/secretsmanager/secretsmanager.go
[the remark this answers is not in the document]
maxres-fr: yea that was a really fustrating thing to work through, especially with new to go errors.

maxres-fr:
> I think we'll want to add to the validation to ensure azureVaultName is set if that secretmanager is chosen. It could be added to this switch statement: https://github.com/ForgeRock/secret-agent/blob/master/api/v1alpha1/api_validation.go#L30

Done.

api/v1alpha1/secretagentconfiguration_types.go
```diff
@@ -123,6 +124,7 @@ type AppConfig struct {
 	SecretsManager          SecretsManager `json:"secretsManager" validate:"required,oneof=none GCP AWS"`
```
jrcast, line 124: need to add Azure in the `oneof` list. `"required,oneof=none GCP AWS Azure"`
Also:
1. Add an entry for Azure in SecretAgentConfigurationStatus #L36 (still not populated, but will)
1. Add azure to CRD field validation #L76 `// +kubebuilder:validation:Enum=none;GCP;AWS,Azure`

pkg/secretsmanager/secretsmanager.go
```diff
@@ -30,6 +34,15 @@ func LoadExisting(ctx context.Context, config *v1alpha1.SecretAgentConfiguration
...
+		if err != nil {
+			return err
+		}
+		err = loadAzureSecrets(ctx, client, config.AppConfig.AzureVaultName, nodes)
+		if err != nil {
+			return err
+		}
 	}
 
 	return nil
```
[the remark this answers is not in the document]
maxres-fr: Yes, I plan that as a second PR

pkg/secretsmanager/secretsmanager.go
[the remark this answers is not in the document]
maxres-fr: `nil` because the caller of the method is going to make a secret if it's `node.Value == nil`
 the really short is to match the GCPLoadSecret https://github.com/ForgeRock/secret-agent/blob/a8157c6edf1f4e7a1c6061cf03fc81cccc643219/pkg/secretsmanager/secretsmanager.go#L173

pkg/secretsmanager/secretsmanager_integration_test.go
```diff
@@ -148,3 +170,83 @@ func TestEnsureGCPSecrets(t *testing.T) {
...
+
+	// create a secret to load
+	ctx2, cancel := context.WithTimeout(baseCtx, 300000000)
+	defer cancel()
+	secretParams := keyvault.SecretSetParameters{}
+	secretName := getSecretID(node.SecretConfig.Namespace, node.Path)
+	secretValue := "test-value"
+	secretParams.Value = &secretValue
+	vaultURL := fmt.Sprintf(azureVaultURLFmt, azureVaultName)
+	_, err = client.SetSecret(ctx2, vaultURL, secretName, secretParams)
```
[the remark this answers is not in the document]
maxres-fr: after it's added.

pkg/secretsmanager/secretsmanager.go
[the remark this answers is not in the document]
jrcast: I guess I missed the title of the PR 🤦‍♂️ : _load existing secrets from Azure key vault_

Issue #22, opened 2020-05-25
Change wenhook_cabundler.go client to use controller-runtime


jrcast:
This file is currently using:
```
	"k8s.io/client-go/kubernetes"
	"k8s.io/client-go/tools/clientcmd"
```
Needs to be updated to use controller-runtime's client as the rest of the project.

snarlysodboxer:
Good plan.

PR #24, opened 2020-06-02
Cleanup code base

### LINKED ISSUE #21: Clean up code base
As we continue to implement and/or clean out kubebuilder's boilerplate code, some existing files become irrelevant. Need to clean out any unused files and make required file structure changes to make the code more cohesive.

### LINKED ISSUE #23: Clean out main.go
In the spirit of keeping the code clean, we should move the webhook-related patching outside main.go into its own file under `controller/`

See the comments for main.go in #16 for details

jrcast:
1. Delete unused files. Fixes #21 
1. Move webhook's cert generation out of main.go. Fixes #23 

other misc fixes:
1. remove unique items from list of objects
2. change default path for keytool and openssl. Use $PATH by default
3. Change base image to  openjdk:11-jre-slim. Half the size of jdk
4. Fix sample config
5. Assigned more memory to the operator


controllers/webhook_cabundler.go
```diff
@@ -12,6 +18,64 @@ type k8s struct {
 	clientset kubernetes.Interface
 }
 
+//InitWebhookCertificates creates and injects req certs by the k8s webhooks
```
snarlysodboxer, line 21:
```suggestion
// InitWebhookCertificates creates and injects req certs by the k8s webhooks
```

controllers/webhook_cabundler.go
```diff
@@ -12,6 +18,64 @@ type k8s struct {
...
+	mutatingWebhookConfigurationName := os.Getenv("MUTATING_WEBHOOK_CONFIGURATION")
+	val := os.Getenv("CERTIFICATE_SANS")
+	dnsNames := strings.Split(val, ",")
+
+	if len(secretName) == 0 || len(namespace) == 0 || len(validatingWebhookConfigurationName) == 0 ||
+		len(mutatingWebhookConfigurationName) == 0 || len(dnsNames) == 0 {
+		os.Exit(1)
+		return errors.New("Need ENVS: WEBHOOK_SECRET_NAME, SERVICE_NAMESPACE, " +
+			"VALIDATING_WEBHOOK_CONFIGURATION, MUTATING_WEBHOOK_CONFIGURATION, CERTIFICATE_SANS")
+	}
```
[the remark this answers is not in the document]
snarlysodboxer: I took me a while to realize you can define flags in a package like this one, and they'll still show up when running the main program.

controllers/webhook_cabundler.go
```diff
@@ -12,6 +18,64 @@ type k8s struct {
...
+		return err
+	}
+	// Unable to create tls.crt
+	if err := ioutil.WriteFile(filepath.Join(certDir, "tls.crt"), cert, 0400); err != nil {
+		return err
+	}
+	// Unable to create tls.key
+	if err := ioutil.WriteFile(filepath.Join(certDir, "tls.key"), key, 0400); err != nil {
+		return err
+	}
```
snarlysodboxer, line 75: What is the reason we're writing these to disk when we're already adding them to k8s secrets?

controllers/webhook_cabundler.go
```diff
@@ -12,6 +18,64 @@ type k8s struct {
 	clientset kubernetes.Interface
 }
```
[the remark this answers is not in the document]
jrcast: Yes. That's coming in #22

controllers/webhook_cabundler.go
jrcast: 🤦

controllers/webhook_cabundler.go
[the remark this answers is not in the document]
jrcast: I think that's worth thinking about. However, I think we would want to do that as a separate PR since this one is just meant to move things around/clean up code. Can you create a ticket for this?

controllers/webhook_cabundler.go
jrcast: The kubebuilder webhook server is setup to get the certificates it needs from a local path. See the docs on `CertDir`
```
// CertDir is the directory that contains the server key and certificate.
// if not set, webhook server would look up the server key and certificate in
// {TempDir}/k8s-webhook-server/serving-certs. The server key and certificate
// must be named tls.key and tls.crt, respectively.
CertDir string
```

jrcast: CertDir is used to create the controller manager. 
https://github.com/ForgeRock/secret-agent/blob/ec5da3c05e7e7695589b2b1939e58c4075920ebb/main.go#L70

snarlysodboxer: I see. Does that mean we're generating new certs every time the operator starts up? If so, do we want to consider only generating them if they don't exist?

controllers/webhook_cabundler.go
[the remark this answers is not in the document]
snarlysodboxer: done https://github.com/ForgeRock/secret-agent/issues/27

controllers/webhook_cabundler.go
jrcast: yes we generate them every time. We don't really do anything with the secret. we can probably  address this later. I don't think it would change functionality or security. Maybe I am missing something out.

api/v1alpha1/secretagentconfiguration_types.go
```diff
@@ -144,7 +144,6 @@ type SecretConfig struct {
 	Namespace string `json:"-" yaml:"namespace,omitempty"`
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:MinItems=1
-	// +kubebuilder:validation:UniqueItems=true
```
snarlysodboxer, line 147: Since we can't use this, we should add custom validation to ensure the keys are unique. Maybe a separate PR though.

config/manager/manager.yaml
```diff
@@ -48,8 +50,8 @@ spec:
         resources:
           limits:
             cpu: 100m
-            memory: 30Mi
+            memory: 100Mi
```
snarlysodboxer, line 53: If we're still having issues with speed, we might try giving it more than 1/10th of one CPU.

controllers/webhook_cabundler.go
snarlysodboxer: It will just increase startup time and CPU time unnecessarily and seems a bit dirty. I have created this ticket: https://github.com/ForgeRock/secret-agent/issues/28

snarlysodboxer: I do think we should make the dir 0750, and use something other than /tmp, otherwise I'll bet EntSec will complain.

api/v1alpha1/secretagentconfiguration_types.go
snarlysodboxer: https://github.com/ForgeRock/secret-agent/issues/29

snarlysodboxer:
I don't see the changes to the k8s manifest to give the app absolute paths to keytool and openssl..  Did I miss them?

controllers/webhook_cabundler.go
jrcast: That's just the folder. The files inside are 400.
but sure, we can make it tighter

jrcast: that can be done as part of #28

jrcast:
> I don't see the changes to the k8s manifest to give the app absolute paths to keytool and openssl.. Did I miss them?

they're there. manager.yaml + keytool.go

PR #30, opened 2020-06-05
Allow full test suite to run containerized

snarlysodboxer:
This allows the integration tests to reach out and hit minikube and work.

### DIFF
--- a/api/v1alpha1/suite_test.go
+++ b/api/v1alpha1/suite_test.go
@@ -1,3 +1,5 @@
+// +build integration
+
 /*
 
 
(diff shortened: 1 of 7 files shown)


api/v1alpha1/suite_test.go
```diff
@@ -1,3 +1,5 @@
+// +build integration
```
jrcast, line 1: can we keep these as part of the unit tests? This just tests that you can add the schema to the api

api/v1alpha1/v1alpha1_test.go
```diff
@@ -1,3 +1,5 @@
+// +build integration
+
```
jrcast, line 2: same here. Just testing you can create a CR object successfully

controllers/suite_test.go
```diff
@@ -1,3 +1,5 @@
+// +build integration
```
jrcast, line 1: same here. Just checks that the controller can start

PR #32, opened 2020-06-05
Add git actions config to build and test on PRs

### LINKED ISSUE #31: Automated testing using git actions
Need to add automated build + test for PRs using git actions

jrcast:
Add git actions config to build and test on PRs
Fixes #31

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -1,0 +1,39 @@
+name: Go
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  basic-checks:
+    name: Run basic-checks
+    runs-on: ubuntu-latest
+    container: golang:1.14-buster
+    steps:
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: go-fmt
+      run: go fmt ./...
+
+    - name: Check for uncommited formatting
+      run: if [ -n "$(git status --untracked-files=no --porcelain)" ]; then echo "There are uncommitted changes"; exit 1; fi
+    
+    - name: Run unit tests
+      run: go test ./... -cover
+    
+  integration-tests:
+    name: Run integration-tests
+    needs: basic-checks
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+       
+    - name: Build Test Image
+      run: docker build -t secret-agent-tester -f Dockerfile.testing .
+
+    - name: Test
+      run: docker run -i --rm -v ${GITHUB_WORKSPACE}:/root/go/src/github.com/ForgeRock/secret-agent secret-agent-tester make citest
+
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -9,8 +9,15 @@ jobs:
   basic-checks:
     name: Run basic-checks
     runs-on: ubuntu-latest
-    container: golang:1.14-buster
+    # container: golang:1.14-buster
     steps:
+
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.14
+      id: go
+
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
--- a/Dockerfile.testing
+++ b/Dockerfile.testing
@@ -1,24 +1,27 @@
 # For building gcr.io/forgerock-io/secret-agent-testing:latest
 
-FROM openjdk:11-jdk-slim
+FROM openjdk:11-jre-slim
+
+ARG GO_VERSION="1.14.4"
+ARG GO_PACKAGE_SHA256="aed845e4185a0b2a3c3d5e1d0a35491702c55889192bb9c30e67a3de6849c067"
+
 ENV CGO_ENABLED=0 GOOS=linux GOARCH=amd64
-CMD ["bash"]
-RUN apt-get update && apt-get install -y curl git-core
-RUN curl -O https://dl.google.com/go/go1.13.10.linux-amd64.tar.gz && \
-    SUM=$(sha256sum go1.13.10.linux-amd64.tar.gz | awk '{print $1}') && \
-    if [ "$SUM" != "8a4cbc9f2b95d114c38f6cbe94a45372d48c604b707db2057c787398dfbf8e7f"]; then echo "Failed checksum"; exit 1; fi && \
-    tar xvf go1.13.10.linux-amd64.tar.gz && \
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y curl git-core make && \
+    apt-get clean all
+RUN curl -LO https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz && \
+    SUM=$(sha256sum go${GO_VERSION}.linux-amd64.tar.gz | awk '{print $1}') && \
+    if [ "${SUM}" != "${GO_PACKAGE_SHA256}" ]; then echo "Failed checksum"; exit 1; fi && \
+    tar xf go${GO_VERSION}.linux-amd64.tar.gz && \
     chown -R root:root ./go && \
     mv go /usr/local && \
-    rm go1.13.10.linux-amd64.tar.gz
-ENV PATH="/usr/local/go/bin:${PATH}" GOPATH="/root/go"
+    rm go${GO_VERSION}.linux-amd64.tar.gz
+
+RUN curl -L https://go.kubebuilder.io/dl/2.3.1/${GOOS}/${GOARCH} | tar -xz -C /tmp/ && \
+    mv /tmp/kubebuilder_2.3.1_${GOOS}_${GOARCH} /usr/local/kubebuilder
+
+ENV PATH="/usr/local/go/bin:${PATH}:/usr/local/kubebuilder/bin" GOPATH="/root/go"
+
 WORKDIR /root/go/src/github.com/ForgeRock/secret-agent
-RUN addgroup --gid 11111 forgerock && \
-    adduser --shell /bin/bash --home /home/forgerock --uid 11111 --disabled-password --ingroup root --gecos forgerock forgerock && \
-    chown -R forgerock:root /home/forgerock
 
-RUN os=$(go env GOOS) && \
-    arch=$(go env GOARCH) && \
-    curl -L https://go.kubebuilder.io/dl/2.3.1/${os}/${arch} | tar -xz -C /tmp/ && \
-    mv /tmp/kubebuilder_2.3.1_${os}_${arch} /usr/local/kubebuilder
-
+CMD ["bash"]
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,12 @@ endif
 all: manager
 
 # Run unit and integration tests (backwards compatability)
+citest: int-test
+	git status --untracked-files=no --porcelain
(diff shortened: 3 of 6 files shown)


jrcast:
test comment

jrcast:
I added branch protection on master. After this merge, I will enable status checks protection as well. PRs will require checks to pass before merging.

Dockerfile.testing
```diff
@@ -1,24 +1,21 @@
...
+FROM openjdk:11-jre-slim
 ENV CGO_ENABLED=0 GOOS=linux GOARCH=amd64
-CMD ["bash"]
-RUN apt-get update && apt-get install -y curl git-core
-RUN curl -O https://dl.google.com/go/go1.13.10.linux-amd64.tar.gz && \
-    SUM=$(sha256sum go1.13.10.linux-amd64.tar.gz | awk '{print $1}') && \
+RUN apt-get update && apt-get install -y curl git-core make && apt-get clean all
+RUN curl -LO https://dl.google.com/go/go1.14.2.linux-amd64.tar.gz && \
+    SUM=$(sha256sum go1.14.2.linux-amd64.tar.gz | awk '{print $1}') && \
     if [ "$SUM" != "8a4cbc9f2b95d114c38f6cbe94a45372d48c604b707db2057c787398dfbf8e7f"]; then echo "Failed checksum"; exit 1; fi && \
```
snarlysodboxer, line 8: Seems like the checksum should've changed with a different version of Go? If this succeeded, then something's wrong with our checksum validation logic.

Dockerfile.testing
```diff
@@ -1,24 +1,21 @@
...
     mv go /usr/local && \
-    rm go1.13.10.linux-amd64.tar.gz
-ENV PATH="/usr/local/go/bin:${PATH}" GOPATH="/root/go"
-WORKDIR /root/go/src/github.com/ForgeRock/secret-agent
-RUN addgroup --gid 11111 forgerock && \
-    adduser --shell /bin/bash --home /home/forgerock --uid 11111 --disabled-password --ingroup root --gecos forgerock forgerock && \
-    chown -R forgerock:root /home/forgerock
+    rm go1.14.2.linux-amd64.tar.gz
+
+RUN curl -L https://go.kubebuilder.io/dl/2.3.1/${GOOS}/${GOARCH} | tar -xz -C /tmp/ && \
```
snarlysodboxer, line 14: Are GOOS and GOARCH definitely available as regular env vars? Depending on what they can at the base URL, I could see this possibly failing silently if those are not regular env vars.

Dockerfile.testing
```diff
@@ -1,24 +1,21 @@
...
 
-RUN os=$(go env GOOS) && \
-    arch=$(go env GOARCH) && \
-    curl -L https://go.kubebuilder.io/dl/2.3.1/${os}/${arch} | tar -xz -C /tmp/ && \
-    mv /tmp/kubebuilder_2.3.1_${os}_${arch} /usr/local/kubebuilder
+ENV PATH="/usr/local/go/bin:${PATH}:/usr/local/kubebuilder/bin" GOPATH="/root/go"
 
+WORKDIR /root/go/src/github.com/ForgeRock/secret-agent
+
+CMD ["bash"]
```
snarlysodboxer, line 21: Great updates to this Dockerfile, much better this way.

.github/workflows/go.yaml
```diff
@@ -0,0 +1,49 @@
...
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  basic-checks:
+    name: Run basic-checks
+    runs-on: ubuntu-latest
```
snarlysodboxer, line 11: I would think we'd want to use docker to run the checks and tests, also to avoid having to `setup-go@v2` every time. Is there a way to do this?

I think this describes how? https://help.github.com/en/actions/creating-actions/creating-a-docker-container-action#creating-an-action-metadata-file Could we then point it to our `Dockerfile.testing` file, then run all our `go vet/fmt/test` commands in that using `uses: actions/our-name`?

.github/workflows/go.yaml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+       
+    - name: Build Test Image
+      run: docker build -t secret-agent-tester -f Dockerfile.testing .
+
+    - name: Test
+      run: docker run -i --rm -v ${GITHUB_WORKSPACE}:/root/go/src/github.com/ForgeRock/secret-agent secret-agent-tester make test
```
snarlysodboxer, line 48: If we're going to automate make here, it means we can't read this file to fully know what this action does, we have to go look at the Makefile to find out what this action really does.

More importantly, since `make test` runs generate and manifests, if running those fix any errors, it will hide the fact that the developer forgot to run them, making the tests pass when they shouldn't.

Personally I think the action should separately run `go vet ./...`, then `go test ./... -tags=integration -cover`, and that's it. `go fmt` will only modify files whose modifications will then go unnoticed, and same with `controller-gen`.

snarlysodboxer: Slight modification, `go test` automatically runs `go vet`, so all we need is the one action of running `go test`.

Dockerfile.testing
jrcast: we set them ourselves in the dockerfile 😁 
https://github.com/ForgeRock/secret-agent/blob/3df51a69f186c0138c71c8ca1196b18878aa6b73/Dockerfile.testing#L4

Dockerfile.testing
```diff
@@ -1,24 +1,21 @@
 # For building gcr.io/forgerock-io/secret-agent-testing:latest
 
-FROM openjdk:11-jdk-slim
+FROM openjdk:11-jre-slim
 ENV CGO_ENABLED=0 GOOS=linux GOARCH=amd64
-CMD ["bash"]
-RUN apt-get update && apt-get install -y curl git-core
-RUN curl -O https://dl.google.com/go/go1.13.10.linux-amd64.tar.gz && \
-    SUM=$(sha256sum go1.13.10.linux-amd64.tar.gz | awk '{print $1}') && \
+RUN apt-get update && apt-get install -y curl git-core make && apt-get clean all
```
maxres-fr, line 5: `DEBIAN_FRONTEND=noninteractive` should be set as ENV (if it's no in upstream)
also `--no-install-recommends --yes`

Dockerfile.testing
snarlysodboxer: Doh. Good call. Hah.

Dockerfile.testing
jrcast: I will take a look at the checksum validation and update the PR. There's definitely a problem.
Good catch! 👍

Dockerfile.testing
```diff
@@ -1,24 +1,21 @@
 # For building gcr.io/forgerock-io/secret-agent-testing:latest
```
maxres-fr, line 1: Also thoughts on using one dockerfile with a testing stage?

.github/workflows/go.yaml
```diff
@@ -0,0 +1,49 @@
+name: Go
```
maxres-fr, line 1: is a GH thing, or can with call it secret-agent. Go isn't a great name.

.github/workflows/go.yaml
jrcast: that's a good point. However, I still there's a benefit of running the generate stuff. That way we can catch problems with `//+kubebuilder` tags

However, I do agree with go fmt. We can kill that.

.github/workflows/go.yaml
maxres-fr: also we should use the same image that's use in the container e.g. buster-slim https://github.com/docker-library/openjdk/blob/bce2fa373dc270cccf539a8e31b5f2a432d23738/11/jre/slim/Dockerfile#L1

Dockerfile.testing
snarlysodboxer: The kubebuilder installation is about 285MB. It's need for testing, but not for production. Is there a way to produce two separate images using multi-stage builds?

.github/workflows/go.yaml
maxres-fr: > More importantly, since make test runs generate and manifests, if running those fix any errors, it will hide the fact that the developer forgot to run them, making the tests pass when they shouldn't.

oof good point

Dockerfile.testing
maxres-fr: yea - you can use `docker build --target testing`

.github/workflows/go.yaml
snarlysodboxer: That's also a good point. What if we either A) run the controller-gen commands after go test, or B) use `git status` or similar to ensure nothing has changed after running the generation commands.

Or maybe both A and B.

Dockerfile.testing
snarlysodboxer: Sweet! I didn't know that one. Using one Dockerfile works for me. 👍

.github/workflows/go.yaml
snarlysodboxer: Actually, of the two options, I think B is the only proper solution, A is not good enough. Maybe someone else has another solution.

jrcast: I will experiment with both A and B. 
I think we'll benefit from having a "quick" `go test` before building Dockerfile.testing + integration tests. Yes, technically tests run twice as of now, but will fail quickly if there's an import problem, etc.

Dockerfile.testing
jrcast: fixed. just a small bug in the if statement

.github/workflows/go.yaml
jrcast: I used a template they had for Go. I think we can name it something else

jrcast: actually, I think we should leave this with "go". We can add different ones later on with different names depending on the type of verification they run. 

Anyway, all actions in this project will be related to `secret-agent` so that'd be implicit

.github/workflows/go.yaml
jrcast: 👍

.github/workflows/go.yaml
jrcast: 👍 
Now it checks for any output from `git status` after `go fmt` and kubebuilder related generators.

Dockerfile.testing
jrcast: I don't think they're exactly compatible. The release Dockerfile is built with a separate stage that builds the go executable then copies that to the final stage.  The testing image does not really need the built package. 

Also, this wont work if we want the standard docker build to be the "release" build. `--target` stops the build when it reaches a specific build stage.  The testing image needs go binaries,  kubebuilder, git, etc (one extra stage) . So the "release" build is the one that will require `--target=release`

`docker build --target release -t myimage`  <<< builds the release image
`docker build -t myimage`  <<< builds the test image

snarlysodboxer: I tend to agree it would be a bit awkward have to build the non-testing image with the --target flag. Maybe @maxres-fr has an idea, but otherwise I see the point in having two Dockerfiles.

jrcast: @maxres-fr WDYT?

maxres-fr: Ok we talked about this on the :phone: we are going to leave as is for now, since the testing stage would be built and discarded during the build process for a release. Moving the release stage to the bottom of the file would make it the default target. The only advantage to the stage being left in would be running tests once again before completing the build.

jrcast: I created #36 to track this. I have a sample Dockerfile. It may/may not be a good idea. But let's see.

PR #39, opened 2020-06-12
CLOUD-2190: generate certs for opendj

maxres-fr:
- Generates certs for opendj encryption
- Fixes issue with lingering keystore inside container
- Generate passwords for opendj

### DIFF
--- a/api/v1alpha1/secretagentconfiguration_types.go
+++ b/api/v1alpha1/secretagentconfiguration_types.go
@@ -173,6 +173,8 @@ type AliasConfig struct {
 	CommonName     string          `json:"commonName,omitempty" yaml:"commonName,omitempty"`
 	Sans           []string        `json:"sans,omitempty" yaml:"sans,omitempty"`
 	SignedWithPath []string        `json:"signedWithPath,omitempty" yaml:"signedWithPath,omitempty"`
+	SelfSigned     bool            `json:"selfSigned,omitempty" yaml:"selfSigned,omitempty"`
+	Expire         int             `json:"expire,omitempty" yaml:"expire,omitempty"`
 	CAPath         []string        `json:"caPath,omitempty" yaml:"caPath,omitempty"`
 	Node           *Node           `json:"-" yaml:"node,omitempty"`
 }
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: forgerock/secret-agent
-  newTag: master
+  newName: registry.local:5000/controller
+  newTag: 11fdfb136e
--- a/pkg/generator/certificate.go
+++ b/pkg/generator/certificate.go
@@ -166,8 +166,47 @@ func GenerateSignedCert(rootCA *Certificate, algorithm v1alpha1.Algorithm, commo
 	if err != nil {
 		return cert, errors.WithStack(err)
 	}
-
 	return cert, nil
+}
+
+// GenerateSelfSignedCertPEM generates a self signed certificate with an expire date based on a unix timestamp
+func GenerateSelfSignedCertPEM(commonName string, expire int) ([]byte, []byte, []byte, error) {
+	var err error
+	cert := &Certificate{}
+	cert.PrivateKeyRSA, err = rsa.GenerateKey(rand.Reader, 3072)
+	if err != nil {
+		return []byte{}, []byte{}, []byte{}, errors.WithStack(errors.New("Unable to generate RSA key"))
+	}
+	marshaledPrivateKey := x509.MarshalPKCS1PrivateKey(cert.PrivateKeyRSA)
+	cert.PrivateKeyPEM = pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY", Bytes: marshaledPrivateKey})
+	// prepare cert template
+	notBefore := time.Unix(0, 0)
+	notAfter := time.Unix(int64(expire), 0)
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
+	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
+	if err != nil {
+		return []byte{}, []byte{}, []byte{}, errors.WithStack(errors.New("Unable to create serial number for cert"))
+	}
+	certTemplate := &x509.Certificate{
+		SerialNumber:          serialNumber,
+		NotBefore:             notBefore,
+		NotAfter:              notAfter,
+		BasicConstraintsValid: true,
+		IsCA:                  false,
+		SignatureAlgorithm:    x509.SHA256WithRSA,
+	}
+	if len(commonName) != 0 {
+		certTemplate.Subject = pkix.Name{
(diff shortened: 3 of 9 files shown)


api/v1alpha1/secretagentconfiguration_types.go
```diff
@@ -173,6 +173,7 @@ type AliasConfig struct {
 	CommonName     string          `json:"commonName,omitempty" yaml:"commonName,omitempty"`
 	Sans           []string        `json:"sans,omitempty" yaml:"sans,omitempty"`
 	SignedWithPath []string        `json:"signedWithPath,omitempty" yaml:"signedWithPath,omitempty"`
+	SharedCert     bool            `json:"sharedCert,omitempty" yaml:"sharedCert,omitempty"`
```
jrcast, line 176: So this bool is to generate an "invalid" cert used by DS. Not sure if "SharedCert" is a good name. Any ideas?

pkg/generator/certificate.go
```diff
@@ -166,10 +166,49 @@ func GenerateSignedCert(rootCA *Certificate, algorithm v1alpha1.Algorithm, commo
 	if err != nil {
 		return cert, errors.WithStack(err)
 	}
-
 	return cert, nil
 }
 
+// GenerateSharedCertPEM generates a "non-conforming" certificate that's intended to not be used in PKI or any application unless explicitly designed to share keys.
+func GenerateSharedCertPEM(commonName string) ([]byte, []byte, []byte, error) {
```
jrcast, line 173: If the only difference is the date and the algo (RSA vs ECDSA) should be pass those as arguments to the function? We can set a default as ECDSA and 10years and you can pass "RSA" and time(0) in this special case.

pkg/generator/generator.go
```diff
@@ -132,44 +132,58 @@ func Generate(node *v1alpha1.Node) error {
...
+				} else {
+					return errors.New("Cert should be either SharedCert or Signed")
 				}
 
 				// add to keystore
 				storePassword, err := getValueFromParent(node.KeyConfig.StorePassPath, node)
 				if err != nil {
 					return err
 				}
+				fmt.Printf("----------%s----------\n", node.AliasConfig.CommonName)
```
jrcast, line 184: debug logs? need to use logging libs instead

maxres-fr: no that's something I meant to remove.

api/v1alpha1/secretagentconfiguration_types.go
maxres-fr: I'm calling it shared, because it's the same cert mounted on all DS instances that's used to encrypt things between DS instances for sharing encrypted things... so?

pkg/generator/certificate.go
maxres-fr: it's also not signed by the CA.

api/v1alpha1/secretagentconfiguration_types.go
jrcast: that's a ds specific use. I thought we wanted to make this generic.

pkg/generator/certificate.go
jrcast: got it

api/v1alpha1/secretagentconfiguration_types.go
jrcast: `placeholderCert`
`invalidCert`
`expiredCert`
`limitedCert`

IDK.. just rambling names that come to mind 😕

maxres-fr: I mean it is kind of. if we wanted to switch the bool, it would would be something like `selfsigned: true` `expire: 1971`. Its just a cert to encrypt shared things and shouldnt be used as PKI which is implied as usable for selfsigned. We can add those it just become more parameters that need to be passed around.  Maybe we could call it `EncryptionCert`

maxres-fr: `NonConformingCert`?

jrcast: If we want to make it 100% generic we can add `algorithm`,  `selfSigned`, and `expire` to `KeyConfig` . That would allow to get rid of the new function as well. Just need to pass these params to the  GenerateSignedCertPEM

pkg/generator/certificate.go
```diff
@@ -166,10 +166,49 @@ func GenerateSignedCert(rootCA *Certificate, algorithm v1alpha1.Algorithm, commo
...
+		return []byte{}, []byte{}, []byte{}, errors.WithStack(err)
+	}
+	cert.CertPEM = pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: certBytes})
+	certAndKeyPEM := append(cert.CertPEM, cert.PrivateKeyPEM...)
+	return certAndKeyPEM, cert.CertPEM, cert.PrivateKeyPEM, nil
+}
+
 // GenerateSignedCertPEM issues a certificate signed by the provided root CA
 //   receives and returns PEM version
 func GenerateSignedCertPEM(rootCAPEM, rootCAPrivateKeyPEM []byte, algorithm v1alpha1.Algorithm, commonName string, sans []string) ([]byte, []byte, []byte, error) {
```
jrcast, line 214: algorithm is already an argument. Would only need to add the expiration time. We can even overload rootCAPem for self-signed. if rootCAPEM is `nil` or any other known value, then self sign. etc the cert.
WDYT?

maxres-fr: I think I'd rather add a `selfSigned` and Expirey property and keep them seperate. Adding a lot of conditionals makes testing a lot harder and creating a bunch of conditional branching is not a thing I'm a huge fan of. `GenerateSelfSigned` would be better with an alg and expirey is a bit more straight forward and should be lessy buggy.

jrcast: sure. Keeping them separate works 👍

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: forgerock/secret-agent
-  newTag: master
+  newName: registry.local:5000/controller
```
jrcast, line 7: we shouldn't push this. 
We want the default to be the public image.

pkg/generator/certificate.go
```diff
@@ -166,10 +166,49 @@ func GenerateSignedCert(rootCA *Certificate, algorithm v1alpha1.Algorithm, commo
 	if err != nil {
 		return cert, errors.WithStack(err)
 	}
-
 	return cert, nil
 }
 
+// GenerateSelfSignedCertPEM generates a self signed certificate with an expire date based on a unix timestamp
+func GenerateSelfSignedCertPEM(commonName string, expire int) ([]byte, []byte, []byte, error) {
```
snarlysodboxer, line 173: What is the difference between this function and `GenerateRootCA` besides PEM encoding everything? The root CA that `GenerateRootCA` creates is a self-signed cert. Maybe instead of reimplementing generation, this function could just call `GenerateRootCA` then PEM encode the things?

snarlysodboxer: And maybe we rename the function to `GenerateRootCAPEM`?

PR #47, opened 2020-06-23
fix: Fix ENABLE_WEBHOOKS flag

jrcast:
if ENABLE_WEBHOOKS==false skip all webhook initialization and cert gen

### DIFF
--- a/main.go
+++ b/main.go
@@ -80,18 +80,18 @@ func main() {
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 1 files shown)


jrcast:
I agree. However, I think we should match kubebuilder's documentation with hopes to keep things familiar to users/devs who use the same framework. See [this doc](https://kubebuilder.io/cronjob-tutorial/running.html#running-and-deploying-the-controller)

PR #52, opened 2020-06-30
Refactor operator types

### LINKED ISSUE #29: Use custom validation to ensure SecretConfig-≥Keys->Name are unique
For more information: https://github.com/ForgeRock/secret-agent/pull/24/files#r435633990

We need to unsure Key Names for SecretConfigs are unique, otherwise one will overwrite another.

jrcast:
1. Complete refactor of the operator types as per new design.
1. Update the validating webhook to account for new data types
1. Implement a mutating webhook to inject default values
1. Add new unit tests and update old ones to account for new data types
1. Add new structure validation criteria. (Fixes #29)


api/v1alpha1/secretagentconfiguration_types.go
```diff
@@ -74,96 +74,159 @@ func init() {
...
 }
 
+// KeySpec is the configuration for each key
+type KeySpec struct {
+	Value                 string        `json:"value,omitempty"`
+	Algorithm             AlgorithmType `json:"algorithm,omitempty"`
+	CommonName            string        `json:"commonName,omitempty"`
+	SignedWithPath        string        `json:"signedWithPath,omitempty"`
+	StoreType             StoreType     `json:"storeType,omitempty"`
+	StorePassPath         string        `json:"storePassPath,omitempty"`
```
wstrange, line 201: I dont think StorePassPath is needed per key - just per keystore
Also- I think we do not use a unique keypass per key - so this could be optionally omitted

api/v1alpha1/secretagentconfiguration_types.go
```diff
@@ -191,3 +254,40 @@ type Node struct {
 	AliasConfig  *AliasConfig  `yaml:"aliasConfig,omitempty"`
 	Value        []byte        `yaml:"value,omitempty"`
 }
+
+// Key Config Type Strings
+const (
+	TypeLiteral      KeyConfigType = "literal"
+	TypePassword     KeyConfigType = "password"
```
wstrange, line 261: Do these constants overlap with the ones defined above?

api/v1alpha1/secretagentconfiguration_types.go
```diff
@@ -191,3 +254,40 @@ type Node struct {
...
+	TypeAESKey      AliasConfigType = "aesKey"
+)
+
+// Algorithm Specifies which keystore algorithm to use
+// +kubebuilder:validation:Enum=ECDSAWithSHA256;SHA256WithRSA
+type Algorithm string
+
+// Algorithm strings
+const (
+	ECDSAWithSHA256 Algorithm = "ECDSAWithSHA256"
```
wstrange, line 291: I guess this is OK - but I'd almost prefer the algo is a free format string that can be passed down to the command? 

i.e. how can this be extended with a new algo without having to recompile and update the CRD

jrcast: Wouldn't we require to update the operator code to support the new algorithm anyway? We can just update the `Enum` at that point. ;) Yes, this would change the CRD behind the scenes, but wouldn't hurt current users.

I personally like that validation of the field happens at the API level and we don't have to code that up in the webhook.

PR #62, opened 2020-07-21
Add owner reference and update status

### LINKED ISSUE #61: Bring back owner reference + sac status
1. generated secrets do not have owner reference
1. Update the secret agent configuration status after reconciler loop

jrcast:
1. Add owner reference to generated secrets
1. Update the SAC status when the reconcile loop is done

closes #61


controllers/secretagentconfiguration_controller.go
```diff
@@ -189,6 +196,29 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
 				"secret_name", secretReq.Name)
 		}
 	}
+	// Update the SecretAgentConfiguration status with the object names
```
maxres-fr, line 199: Any thoughts on this being its own func that updates stats?

jrcast: done

PR #68, opened 2020-07-23
Limit reconcile runs, reduce logs, remove ENVs

### LINKED ISSUE #27: Unify configuration methodologies
In `controllers/webhook_cabundler.go` we've setup environment variables to configure the app. I believe everywhere else we're using the `flag` package.

Do we want to consider switching these to flag configs as well? It also provides an easy way to set sane defaults where possible, avoiding the requirement of having to think about these things in order to run the app.

Global vars, that are checked in an init function if sane defaults can't be provided make sense to me.

More info: https://github.com/ForgeRock/secret-agent/pull/24/files#r435334765

### LINKED ISSUE #66: cleanup/reduce  logs
The operator prints out too many logs. We should convert these logs to debug logs

### LINKED ISSUE #67: limit the number of reconcile calls
The reconcile loop is called every time owned secrets and a `sac` is modified/created. Since we update the k8s secrets for each key, we end up running the reconcile loop dozens of times. Need to update the secrets only when necessary to limit the number of reconcile loop runs.

jrcast:
1. Reduce number of reconcile runs. Fixes #67
1. Reduce logs printed. Converted some logs to debug. Fixes #66
1. Use flags for all arguments. Removed ENVs. Fixes #27

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -30,19 +30,13 @@ spec:
         - --enable-leader-election
         - --keytoolPath=/usr/local/openjdk-11/bin/keytool
         - --opensslPath=/usr/bin/openssl
+        - --webhook-service-ns=$(SERVICE_NAMESPACE)
+        - --webhook-service-name=$(SERVICE_NAME)
+        - --webhook-secret-name=$(WEBHOOK_SECRET_NAME)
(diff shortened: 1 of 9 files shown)


maxres-fr:
Awesome, so happy that you were able to just prevent the watch during
updates.

One thing:

We need to make sure there's no `fatal()` calls defer won't run. That could
leave us with unwatched.  Maybe we need to check for that on start?

PR #69, opened 2020-07-23
Keytool object

maxres-fr:


pkg/generator/rootca.go
```diff
@@ -38,7 +38,7 @@ func (rCA *RootCA) References() ([]string, []string) {
 }
 
 // LoadReferenceData loads references from data
-func (rCA *RootCA) LoadReferenceData(data []map[string][]byte) error {
+func (rCA *RootCA) LoadReferenceData(data map[string][]byte) error {
```
jrcast, line 41: This would have to be updated in all keyTypes. I think this should happen as part of this PR

controllers/secretagentconfiguration_controller.go
```diff
@@ -18,6 +18,7 @@ package controllers
```
jrcast, line 18: you may be better off applying your changes to the file manually rather than trying to merge

pkg/generator/rootca.go
maxres-fr: @jrcast :+1: I need to update the type you merged. 

- [ ] update password type
- [ ] update literal type

controllers/secretagentconfiguration_controller.go
```diff
@@ -121,12 +119,13 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
 					"secret_type", string(key.Type))
 				keyInterface.LoadFromData(secObject.Data)
 			}
-
+			// TODO maybe we don't need to use the slice?
 			// Load key references and data
-			var keyRefSecrets []map[string][]byte
+			var keyRefSecrets map[string][]byte
```
jrcast, line 124: You need to allocate the map before assigning values to it
```suggestion
			keyRefSecrets := make(map[string][]byte)
```

maxres-fr: doh

pkg/generator/keytool.go
```diff
@@ -25,6 +27,133 @@ func init() {
...
+
+// AliasMgr an interface for managing keytool aliases
+type AliasMgr interface {
+	References() ([]string, []string)
+	LoadReferenceData(data map[string][]byte) error
+	Generate() error
+}
+
+// NewKeyTool creates new keytool instance
+func NewKeyTool(key *v1alpha1.KeyConfig) *KeyTool {
```
jrcast, line 51: I think this should return (*KeyTool, error) to keep things consistent with other generators
```suggestion
func NewKeyTool(key *v1alpha1.KeyConfig) (*KeyTool, error) {
```

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: forgerock/secret-agent
-  newTag: master
+  newName: controller
```
maxres-fr, line 7: this shouldn't be committed.

pkg/generator/keytoolimportpassoword.go
```diff
@@ -0,0 +1,64 @@
+package generator
```
maxres-fr, line 1: file name isn't spelled correctly.

PR #81, opened 2020-07-31
Add retry with exponential backoff on recoverable error

### LINKED ISSUE #71: We should retry the reconcile loop if we fail to create a secret
given secret dependencies, we may require more than one iteration of the reconcile loop to create all secrets. We should trigger a retry if we fail to create a secret. We should only retry a MAX number of times to avoid an infinite loop. Maybe 10?

jrcast:
1. Add retry with exponential backoff on recoverable error
1. Add new `State` to SAC Status
1. Update client-go module

Closes #71


controllers/secretagentconfiguration_controller.go
```diff
@@ -229,6 +243,14 @@ func (reconciler *SecretAgentConfigurationReconciler) updateStatus(ctx context.C
 	instance.Status.ManagedK8sSecrets = secretNames
 	instance.Status.TotalManagedObjects = totalManagedObjects
 
+	if errorFound {
+		instance.Status.State = v1alpha1.SecretAgentConfigurationError
+	} else if inProgress {
+		instance.Status.State = v1alpha1.SecretAgentConfigurationInProgress
+	} else {
```
maxres-fr, line 250: We default `errorFound` to false, if we miss catching some error or whatever in the reconcile loop this leads to a default state of "completed". I'd rather have an unknown or unhandled state as a safety, I think the "complete" state might lead to false positives.

go.mod
```diff
@@ -12,16 +12,16 @@ require (
...
 	github.com/go-logr/logr v0.1.0
 	github.com/go-playground/validator/v10 v10.2.0
-	github.com/onsi/ginkgo v1.11.0
-	github.com/onsi/gomega v1.8.1
+	github.com/onsi/ginkgo v1.12.1
+	github.com/onsi/gomega v1.10.1
 	github.com/pkg/errors v0.9.1
 	go.uber.org/zap v1.10.0
 	golang.org/x/crypto v0.0.0-20200510223506-06a226fb4e37
+	golang.org/x/time v0.0.0-20191024005414-555d28b269f0
```
maxres-fr, line 20: time isn't part of stdlib? _jusy say'n?_

api/v1alpha1/secretagentconfiguration_types.go
```diff
@@ -70,6 +71,16 @@ type SecretAgentConfigurationList struct {
 	Items           []SecretAgentConfiguration `json:"items"`
 }
 
+// SecretAgentConfState is used to keep track of the SAC state
+type SecretAgentConfState string
+
+// SecretsManager Strings
+const (
```
maxres-fr, line 78: See my comment in the updatestatus method re: statuses.

api/v1alpha1/secretagentconfiguration_types.go
```diff
@@ -70,6 +71,16 @@ type SecretAgentConfigurationList struct {
...
+	SecretAgentConfigurationCompleted  SecretAgentConfState = "Completed"
+	SecretAgentConfigurationInProgress SecretAgentConfState = "InProgress"
+	SecretAgentConfigurationError      SecretAgentConfState = "ErrorFound"
+)
```
maxres-fr, line 82:
```suggestion
const (
	SecretAgentConfigurationCompleted  SecretAgentConfState = "Completed"
	SecretAgentConfigurationInProgress SecretAgentConfState = "InProgress"
	SecretAgentConfigurationError      SecretAgentConfState = "ErrorFound"
	SecretAgentConfigurationMisconfigured      SecretAgentConfState = "Misconfigured"
)
```

controllers/secretagentconfiguration_controller.go
maxres-fr:
```suggestion
	} else if !errorFound && !inProggress {
			instance.Status.State = v1alpha1.SecretAgentConfigurationInProgress
	} else {
	instance.Status.State = v1alpah1.SecretAgentConfigurationMisconfigured
	}
```

jrcast: we decided to keep it as is

PR #88, opened 2020-08-11
controller stores reference secrets using composite of name and data key

### LINKED ISSUE #84: Controller and LoadReferences assume the data key is uniq across all secrets
The controller and keymgr interfaces need to expect that during loadrefernce is called the reference data around https://github.com/ForgeRock/secret-agent/blob/refactor/controllers/secretagentconfiguration_controller.go#L129 will be merged `SecretName/KeyName` or whatever is returned from `References`. This is to prevent collisions with secrets that have the same keyname

maxres-fr:
Controller stores reference data using both slices returned from
References(). The keyname pattern is "slice1[i]/slice2[i]". For most
KeyMgrs this will match the pattern used in the Spec. Root Certs and SSH
keys are different and use specific names always.

KeyMgr.LoadReferenceData(data) should now expect to data to be keyed "slice1[i]/slice2[i]"

closes #84

### DIFF
--- a/controllers/secretagentconfiguration_controller.go
+++ b/controllers/secretagentconfiguration_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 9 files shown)


jrcast:
LGTM

PR #92, opened 2020-08-13
Add ability to use binary characters in passwords

### LINKED ISSUE #89: Password Generation Type needs security enhancements
Currently passwords are generated using only alpha numeric values. This should include all printable e.g readable ascii characters (Dec 32 - 127)

There's a should be "binary" formatted passphrases can be made up of the full 256 bytes in ascii.

jrcast:
Use binary characters in passwords when `useBinaryCharacters: true`

closes #89

### DIFF
--- a/pkg/generator/password.go
+++ b/pkg/generator/password.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"math/big"
 
+	b64 "encoding/base64"
+
 	corev1 "k8s.io/api/core/v1"
 
 	"github.com/ForgeRock/secret-agent/api/v1alpha1"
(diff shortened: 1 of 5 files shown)


pkg/generator/password.go
```diff
@@ -67,15 +70,29 @@ func (pwd *Password) InSecret(secObject *corev1.Secret) bool {
 
 // Generate generates data
 func (pwd *Password) Generate() error {
-
+	var max *big.Int
 	alphanumericBytes := "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
```
maxres-fr, line 74: I thought we were going to use "symbols" too. e.g. 32 - 127?

PR #98, opened 2020-08-18
use absolute paths for external commands

### LINKED ISSUE #97: use absolute paths for external binaries
use an absolute path to keytool and openssl

maxres-fr:
closes #97


jrcast:
tests are failing. Seems the tester image target does not like the new default path.

maxres-fr:
Yes it's in the JAVA_HOME.

PR #100, opened 2020-08-18
adds ability to self sign and create an expired cert

### LINKED ISSUE #86: self signed support lost
We lost the ability to generate keypairs that are self signed. This is required for at least DS.

maxres-fr:
closes #86

### DIFF
--- a/pkg/generator/certificate.go
+++ b/pkg/generator/certificate.go
@@ -127,9 +127,23 @@ func (kp *CertKeyPair) Generate() error {
 		marshaledPrivateKey := x509.MarshalPKCS1PrivateKey(kp.Cert.PrivateKeyRSA)
 		kp.Cert.PrivateKeyPEM = pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY", Bytes: marshaledPrivateKey})
 	}
+	// setup expire
 	// prepare cert template
+	currentTime := time.Now()
 	notBefore := time.Now().Add(time.Minute * -5)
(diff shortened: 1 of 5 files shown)


pkg/generator/certificate.go
```diff
@@ -127,9 +127,19 @@ func (kp *CertKeyPair) Generate() error {
...
+	// setup expire
 	// prepare cert template
+	currentTime := time.Now()
 	notBefore := time.Now().Add(time.Minute * -5)
-	notAfter := notBefore.Add(10 * 365 * 24 * time.Hour) // 10yrs
+	// default 10yrs
+	notAfter := notBefore.Add(10 * 365 * 24 * time.Hour)
+
+	// forcing an expired cert
+	if currentTime.After(currentTime.Add(kp.V1Spec.Duration.Duration)) {
```
jrcast, line 138: 🤔  ???

maxres-fr: We are supporting the creation of expired certs for DS. DS uses those for encrypt things but never part of anything PKI related, so they are set to be expired just in the event someone attempts to re-use the certs as part of a PKI, then at least most HTTP clients will reject due to being expired. 

The check here is if the duration is set to a negative value e..g `-10y` the addition of the current time and the duration would yield a result being before the current time, therefore the request is for an expired cert. Now that I've written that I see that logic is reversed... probably should wire up some tests :-1:

jrcast: + add some comments 😅

PR #102, opened 2020-08-20
fixes a couple bugs with self signed, and handles some fields that in…

maxres-fr:
… rare instances be nil


maxres-fr:
Long term I don't think this is a good approach to handling cases where random fields are nil/have no value. If we always use the constructor methods to create a new instance instead of using a struct literal would be much safer. This mostly occurs in tests, during operator install and when trying to debug locally without the validating webhook running.

maxres-fr:
2C8F

PR #107, opened 2020-08-25
adds importcert, no usage attributes for negative duration values

### LINKED ISSUE #103: importcert as trustcacerts
In some cases some applications need a CA's public key to be imported outside of the truststore configuration. e.g `ds` needs its keystore to have `ca-cert`.

maxres-fr:
closes #103

### DIFF
--- a/config/samples/secret-agent_v1alpha1_secretagentconfiguration.yaml
+++ b/config/samples/secret-agent_v1alpha1_secretagentconfiguration.yaml
@@ -70,10 +70,14 @@ spec:
         algorithm: SHA256WithRSA
         distinguishedName:
           commonName: master-key
-        signedWithPath: platform-ca/ca
+        selfSigned: true
+        # any negative duration produces a keypair that has 
+        # an unusable valid period
+        duration: "-72h"
+
     - name: keystore.pin
       type: password
-    - name: keystore.p12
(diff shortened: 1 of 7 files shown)


config/samples/secret-agent_v1alpha1_secretagentconfiguration.yaml
```diff
@@ -84,10 +88,10 @@ spec:
             cmd: importkeystore
             sourcePath: "platform-ca/ca"
           - name: ssl-keypair
-            cmd: importkeystore 
+            cmd: importkeystore
             sourcePath: "ds/ssl-key-pair"
           - name: master-keypair
-            cmd: importkeystore  
+            cmd: import
```
jrcast, line 94: should this be `importcert`?
```suggestion
            cmd: importcert
```

maxres-fr: this is `importkeystore`, I've been refining this spec on the forgeops repo and will do a bigger add back to this repo. 

It's importkeystore because the master-key-pair is used for encryption/decryption on DS data.

jrcast: `import` is not defined in the types file :(

PR #113, opened 2020-09-03
Implement truststore secret manager support

### LINKED ISSUE #48: "truststore" secret type breaks secretmanager.EnsureSecrets
Seems there's a problem when processing `truststore` with `caCopy`
using a quick debug print:
```
// add secret version
valueString := string(node.Value)
fmt.Printf("The length of %q is %d\n", secretID, len(valueString)) //DEBUG PRINT
```
printed
```
The length of "default-amster-authorized-keys" is 381
The length of "default-truststore-cacerts" is 2306
The length of "default-truststore-cacerts-ca-cert" is 0 <<< ca-cert seems to be empty at the moment it is EnsureSecrets is called.
```
You can't create secrets with length=0 

removing `truststore` from the `sac` allows the secretmanager to work again.

### LINKED ISSUE #111: Truststore secret manager support
Implement support for truststore in secret manager

jrcast:
Implement truststore secret manager support

Closes #111, Closes #48

### DIFF
--- a/pkg/generator/truststore.go
+++ b/pkg/generator/truststore.go
@@ -4,15 +4,16 @@ import (
 	"context"
 	"crypto/rand"
 	"crypto/x509"
+	"encoding/binary"
 	"encoding/pem"
 	"fmt"
 	"io/ioutil"
 
+	"github.com/ForgeRock/secret-agent/api/v1alpha1"
+	"github.com/ForgeRock/secret-agent/pkg/secretsmanager"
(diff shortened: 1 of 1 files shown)


maxres-fr (changes requested):
we wont be needing to put truststores in a secret manager. They should always be re-generated from source. Truststores are combination of system certificates + any requested certs to be appended. The appended certs are references ergo a reference is the source of truth.

PR #116, opened 2020-09-04
Add prefix for secrets stored in secret manager, add back azure func …

### LINKED ISSUE #115: Secret Manager Cluster ID
We have an issue with secret manager support where we name things with a pattern of 
`<namespace>_<secret name>_<keyname>` which works fine until there's multiple clusters with the same namespace.

We need to find a way to create a uniq id to identify a cluster but it will also need to have a "restorable" value because a user might remove a cluster, then re provision one and expect to be able to deploy with those same secrets again.

maxres-fr:
…for loading from vault

closes #115

### DIFF
--- a/api/v1alpha1/secretagentconfiguration_types.go
+++ b/api/v1alpha1/secretagentconfiguration_types.go
@@ -184,6 +184,8 @@ type AppConfig struct {
 	GCPProjectID          string         `json:"gcpProjectID,omitempty"`
 	AWSRegion             string         `json:"awsRegion,omitempty"`
 	AzureVaultName        string         `json:"azureVaultName,omitempty"`
+	// +kubebuilder:validation:Optional
+	SecretManagerKeyPrefix string `json:"secretManagerKeyPrefix"`
 
 	// Optional number of times the operator will attempt to generate secrets. Defaults to 3
 	MaxRetries *int `json:"maxRetries,omitempty"`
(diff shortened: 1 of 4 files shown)


api/v1alpha1/secretagentconfiguration_types.go
```diff
@@ -184,6 +184,8 @@ type AppConfig struct {
 	GCPProjectID          string         `json:"gcpProjectID,omitempty"`
 	AWSRegion             string         `json:"awsRegion,omitempty"`
 	AzureVaultName        string         `json:"azureVaultName,omitempty"`
+	// +kubebuilder:validation:Optional
+	SecretManagerKeyPrefix string `json:"secretManagerKeyPrefix"`
```
jrcast, line 188: you can just add `omitempty` to match the other fields

pkg/secretsmanager/secretsmanager.go
```diff
@@ -27,6 +28,9 @@ func idSafe(value string) string {
 // EnsureSecret ensures a secret is stored in secret manager
 func EnsureSecret(ctx context.Context, config *v1alpha1.AppConfig, secretName string, value []byte) error {
 	secretID := idSafe(secretName)
+	if config.SecretManagerKeyPrefix != "" {
+		secretID = fmt.Sprintf("%s_%s", config.SecretManagerKeyPrefix, secretID)
```
jrcast, line 32:
```suggestion
		secretID = fmt.Sprintf("%s-%s", config.SecretManagerKeyPrefix, secretID)
```

pkg/secretsmanager/secretsmanager.go
```diff
@@ -49,6 +53,9 @@ func EnsureSecret(ctx context.Context, config *v1alpha1.AppConfig, secretName st
 // LoadSecret Loads secrets from the configured secret manager
 func LoadSecret(ctx context.Context, config *v1alpha1.AppConfig, secretName string) ([]byte, error) {
 	secretID := idSafe(secretName)
+	if config.SecretManagerKeyPrefix == "" {
```
jrcast, line 56:
```suggestion
	if config.SecretManagerKeyPrefix != "" {
```

jrcast (changes requested):

PR #117, opened 2020-09-04
Add secretsManagerPrefix to appConfig in the SAC

### LINKED ISSUE #115: Secret Manager Cluster ID

jrcast:
1. Add secretsManagerPrefix to AppConfig in the SAC
1. Fix bug with AWS creds. Use  secret only if user provides creds for AWS
1. Add  missing azure secret manager function. WIP

Closes #115


maxres-fr (approved):
:tada:

PR #134, opened 2020-10-02
Rsa root cert

maxres-fr:


pkg/generator/certificate.go
```diff
@@ -316,6 +313,7 @@ func (kp *CertKeyPair) Generate() error {
 
 // LoadFromData  load data from kubernetes secret
 func (kp *CertKeyPair) LoadFromData(data map[string][]byte) {
+	// TODO
```
jrcast, line 316: what's there TODO?

pkg/generator/certificate.go
```diff
@@ -339,8 +337,10 @@ func (kp *CertKeyPair) LoadFromData(data map[string][]byte) {
 	switch priv := privateKey.(type) {
 	case *ecdsa.PrivateKey:
 		kp.Cert.PrivateKeyEC = priv
+		kp.V1Spec.Algorithm = v1alpha1.AlgorithmTypeECDSAWithSHA256
```
jrcast, line 340: take a look at https://github.com/ForgeRock/secret-agent/pull/134/files#diff-675de4ffdac7698a359e45836ac5051fL428

rootCA is hardcoded to ECDSA. I think we could make this the default in the mutating webbook if the user does not specify anything. However, let the user specify RSA if they want to. 🤷 ???

jrcast (approved):

pkg/generator/certificate.go
maxres-fr: I added the hook

pkg/generator/certificate.go
```diff
@@ -425,6 +424,8 @@ func NewRootCA(keyConfig *v1alpha1.KeyConfig) (*CertKeyPair, error) {
 		Cert: &Certificate{},
 	}
 	rCA.V1Spec = keyConfig.Spec
-	rCA.V1Spec.Algorithm = v1alpha1.AlgorithmTypeECDSAWithSHA256
+	if rCA.V1Spec.Algorithm == "" {
+		rCA.V1Spec.Algorithm = v1alpha1.AlgorithmTypeECDSAWithSHA256
+	}
```
jrcast, line 429: We already handle this in the webhook change above. The algorithm will already be set by the time we get here.
```suggestion
```

jrcast (approved):

pkg/generator/certificate.go
maxres-fr: Yep - I did that on purpose. we might get a null pointer on this if the webhooks get skipped

jrcast: it shouldn't 🤔 unless you're running debugging/running locally.

PR #135, opened 2020-10-02
Controller improvements

### LINKED ISSUE #130: Race condition: create each k8s secret after we finish processing all its keys
We create and update k8s secrets as we process each key in the SAC. We should manage the k8s secrets after we're done processing all its keys. 

As of now, we can hit a race condition in which pods start using/mounting the secrets even before they are finished processing. If any pod is scheduled to run before the secret is complete, it will have missing values. This can be hard to debug as the secrets will eventually have all required keys but the pod wont.

### LINKED ISSUE #131: Add "state" to the kubectl get view
We should include the `state` field in the output of `kubectl get sac`. 

```
jrcast@lmde4:~/dev/go/src/github.com/ForgeRock/secret-agent/demo$ k -n test-sa get sac
NAME            TOTALNUMOBJECTS
forgerock-sac   12
```

```
kubectl describe sac abc
...
...
...

Status:
  managedK8sSecrets:
    truststore
    ds-passwords
    am-keystore
    platform-ca
    am-env-secrets
    ds-env-secrets
    idm-env-secrets
    idm
    am-passwords
    amster-env-secrets
    ds
    amster
  State:                  Completed  <<<<<<<<<<<<<<This value
  Total Managed Objects:  12
Events:                   <none>
```

jrcast:
1. Add state to the additionalPrinterColumns with the state of the SACs  e.a. `kubectl get sac --all`
1. Write secret to k8s after processing all its keys

Closes #130
Closes #131

### DIFF
--- a/controllers/secretagentconfiguration_controller.go
+++ b/controllers/secretagentconfiguration_controller.go
@@ -119,6 +119,13 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
 		}
 
 	}
+	// set the SAC status to inProgress only the first time around.
+	if instance.Status.State == "" {
+		if err := reconciler.updateStatus(ctx, &instance, true, false); err != nil {
+			log.Error(err, "Failed to update status", "instance.name", instance.Name)
+			return ctrl.Result{}, err
+		}
+	}
 
 	for _, secretReq := range instance.Spec.Secrets {
 		// load from secret k8s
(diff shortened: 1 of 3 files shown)


controllers/secretagentconfiguration_controller.go
```diff
@@ -179,20 +187,26 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
...
-				secRefObject, err := k8ssecrets.LoadSecret(reconciler.Client, ref, instance.Namespace)
-
-				if err != nil {
-					log.Error(err, "error looking up secret ref",
-						"secret_name", secretReq.Name,
-						"secret_ref", ref)
-					rescheduleRetry, errorFound = true, true
+				var secRefObject *corev1.Secret
+				// if ref the current secret, use the values already loaded in memory, else load them from k8s
+				if ref == secObject.Name {
```
maxres-fr, line 192: I think most of the block around LOC 205 can be in this if statement, and use the break here. 

This should let you remove the else LOC 194 and make it a little more concise.

maxres-fr (approved):

PR #136, opened 2020-10-06
Ability to select the source NS for cloud creds

### LINKED ISSUE #133: credentialsSecretName should reference secrets in the operator's namespace
Currently, `credentialsSecretName` references secrets in the same namespace as the SAC. This means the cloud credentials need to be present in the same namespace as the SAC.  

Need to add support to reference secrets located elsewhere.

jrcast:
1. Users can force the NS the operator uses to source cloud creds
1. The NS is provided using ` --cloud-secrets-namespace` runtime arg.
1. If the arg is omitted, it assumes the creds NS is the same as the SAC
1. The sample deployment uses the same NS as the operator by default
1. Update the README.md to account for these changes.

Closes #133

### DIFF
--- a/controllers/secretagentconfiguration_controller.go
+++ b/controllers/secretagentconfiguration_controller.go
@@ -45,8 +45,9 @@ import (
 // SecretAgentConfigurationReconciler reconciles a SecretAgentConfiguration object
 type SecretAgentConfigurationReconciler struct {
 	client.Client
-	Log    logr.Logger
-	Scheme *runtime.Scheme
+	Log                   logr.Logger
+	Scheme                *runtime.Scheme
+	CloudSecretsNamespace string
 }
 
 var (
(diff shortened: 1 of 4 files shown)


controllers/secretagentconfiguration_controller.go
```diff
@@ -84,14 +85,21 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
 
 	if instance.Spec.AppConfig.SecretsManager != v1alpha1.SecretsManagerNone {
 		var dir string
+		var cloudCredNs string
```
maxres-fr, line 88: kind of a nitpick but I think this should be `cloudCredNS` surprised go fmt/lint didn't catch that.

maxres-fr (approved):

PR #138, opened 2020-10-13
Implement test for `keytool -importcert` cmd

### LINKED ISSUE #106: importcerts tests

jrcast:
1. Implement test for `keytool -importcert` cmd
1. Rename importpassword_test.go to match convention

Closes #106


pkg/generator/keytoolimportcert_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/ForgeRock/secret-agent/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestImportCert(t *testing.T) {
+	rootCAConfig := &v1alpha1.KeyConfig{
```
maxres-fr, line 15: I think at some point we might want to use a helper func for this, I think this is a bit repeated but yea. 

I also wonder if some of this can be converted to test tables.

maxres-fr (approved):

PR #140, opened 2020-10-16
Add support for GCP workload identity

### LINKED ISSUE #132: Investigate use of workload identity on GCP
As a cluster operator, I'd like the secret agent operator to use workload identity to access the GCP Secret Manager

jrcast:
1. Add support for GCP workload identity
1. Update README with new steps to enable workload identity
1. Add GCP secret manager fields to the sample SAC.
1. Make "managed identities" the default authentication method for Azure AKS

Closes #132

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ spec:
 
 _note:_ Azure's API response time on Key Vault is long and will delay the creation of secrets. It might be beneficial to deploy a SAC before long before deploying an application if use Azure Key Vault
 
-The `secret-agent` uses credentials which are available using two different methods: Azure Managed Identities or explicit credentials. Both are configured in the secret that's referenced in the SAC spec `spec.appConfig.credentialsSecretName`. Example Azure Configuration for a SAC:
+The `secret-agent` uses credentials which are available using two different methods: Azure Managed Identities (recommended for Azure deployemnts) or explicit credentials. Explicit credentials are configured in a secret referenced in the SAC spec `spec.appConfig.credentialsSecretName`. Example Azure Configuration for a SAC:
 
 ```yaml
 apiVersion: secret-agent.secrets.forgerock.io/v1alpha1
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -103,13 +103,50 @@ data:
...
+```bash
+PROJECTID=myproject #GCP project ID
+GSA-NAME=mygcpserviceaccount #GCP service account name
+# Create the GCP IAM policy binding
+gcloud iam service-accounts add-iam-policy-binding --role roles/iam.workloadIdentityUser --member "serviceAccount:${PROJECTID}.svc.id.goog[secret-agent-system/secret-agent-manager-service-account]" ${GSA-NAME}@${PROJECTID}.iam.gserviceaccount.com
+# Annotate the Kubernetes service account
+kubectl -n secret-agent-system annotate serviceaccounts secret-agent-manager-service-account iam.gke.io/gcp-service-account=${GSA-NAME}@${PROJECTID}.iam.gserviceaccount.com
+```
+
+**Note: in order to use workload identity, no `spec.appConfig.credentialsSecretName` should be provided**. If credentials are provided, `secret-agent` will use the provided credentials instead.
```
maxres-fr, line 128: I'm wondering if we should switch around Azure logic to behave this way too....

maxres-fr: I do fear that without an explicitly set flag for "use workload id" users could run into and issue where they expect it to work but it doesn't.

README.md
```diff
@@ -103,13 +103,50 @@ data:
...
 
 Please refer to the [GCP Documentation](https://cloud.google.com/secret-manager/docs/reference/libraries?hl=nl#cloud-console) for instructions on how to create a service account with the necessary permissions to access the GCP Secrets Manager. The `secret-agent` needs access to read/write secrets. This can be achieved by assigning the `Secret Manager Admin` role to the service account provided.
 
+##### Workload identity
+
+Workload Identity is the recommended way to access Google Cloud services from applications running within GKE. For more information on how to enable workload identity see [GCP Documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#gcloud_1).
+
+In general, once the Google service account has been created with the proper role attached and workload identity has been enabled for your GKE cluster and nodes, you can run the following commands to enable workload identity for the `secret-agent`:
+
+```bash
```
maxres-fr, line 119: I question if we should offer these instructions other than a pointer to the documents

jrcast: I guess we could. Make the `cloud-credentials` reference optional and remove the references to `AZURE_MANAGED_ID`. Instead of setting that to "true", just skip the credentials file altogether. If no reference is provided, do nothing (like what currently happens if `AZURE_MANAGED_ID` is set to "true")

README.md
jrcast: Not sure about this. Take AWS as an example, we don't have a setting to use instance metadata or not. The workload identify setup happens outside the operator itself so there's nothing we do to turn it on or off.

jrcast: updated README.md to clarify azure's "managed identity" is used if no secret is provided

maxres-fr (approved):

PR #145, opened 2020-10-30
remove useless line in releae target

maxres-fr:


Makefile
```diff
@@ -62,7 +63,7 @@ deploy: manifests
 release:
 	cd config/manager && kustomize edit set image controller=${IMG}
 	kustomize build config/default 
-	GOPROXY=https://proxy.golang.org GO111MODULE=on go get github.com/ForgeRock/secret-agent@v$VERSION
+	GOPROXY=https://proxy.golang.org GO111MODULE=on go get github.com/ForgeRock/secret-agent@$(VERSION)
```
jrcast, line 66: this may be too crazy but should be tag and push tags as part of the `release`?
```suggestion
        git tag $(VERSION)
        git push --tags
	curl https://proxy.golang.org/github.com/forgerock/secret-agent/@v/$(VERSION).info
```

do note that I changed to a simple GET request. According to https://go.dev/about, it should work

jrcast (approved):
LGTM

PR #147, opened 2020-11-05
fix: controller moves to the next secret if a reference doesn't exist

maxres-fr:


wstrange (approved):
thx for the quick turn around

jrcast (approved):
ship it

PR #149, opened 2020-11-12
Handle out-of-order self-references in secrets

### LINKED ISSUE #148: should be able to handle out-of-order self-references in secrets
Secret agent should be able to handle out-of-order self-references. Currently, the `secret-agent` falls into an infinite loop if a key references another key form the same secret before it is defined in the `SAC`.

for example: 
```yaml
  - name: ds
    keys:
    - name: keystore
      # Java Keystore type
      type: keytool
      spec:
        storeType: pkcs12
        storePassPath: ds/keystore.pin # <<< Reference keystore.pin before it has been defined
        keyPassPath: ds/keystore.pin  # <<< Reference keystore.pin before it has been defined
        keytoolAliases:
          - name: configstorepwd
            # import a password/passhphrase
            cmd: importpassword
            sourcePath: "am-passwords/.storepass"
    - name: keystore.pin # <<<<<<  keystore.pin  defined here
      type: password
      spec:
        length: 32
      # keystore used by DS
```

jrcast:
Handle out-of-order self-references in secrets

Closes #148

### DIFF
--- a/controllers/secretagentconfiguration_controller.go
+++ b/controllers/secretagentconfiguration_controller.go
@@ -133,6 +133,7 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
 specSecretsLoop:
 	for _, secretReq := range instance.Spec.Secrets {
 		var k8sApplyRequired bool = false
+		var redoKeys bool = false
 		// load from secret k8s
 		secObject, err := k8ssecrets.LoadSecret(reconciler.Client, secretReq.Name, instance.Namespace)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


wstrange (approved):

controllers/secretagentconfiguration_controller.go
```diff
@@ -274,6 +285,14 @@ specSecretsLoop:
 			// if we reach this point, we need to update the k8s secret with new keys
 			k8sApplyRequired = true
 		} // end of keys loop
+		// if redoKeys is true, rerun the keys loop
+		if redoKeys {
+			log.V(1).Info("retrying all keys from secret",
+				"secret_name", secretReq.Name)
+			redoKeys = false
+			// HOTFIX: This is bad practice. We shoudln't use `goto`. The reconcile loop must be refactored asap.
```
jrcast, line 293: Added this comment to the code. The reconcile loop is now too long and we should split it into functions (or any other fancier solution). See #150

PR #156, opened 2020-12-01
Add finalizers to SAC and secrets

### LINKED ISSUE #154: Add finalizers to the SAC
Add finalizers to the SAC and other objects to reduce race conditions during namespace deletions.

### LINKED ISSUE #155: Fix race condition with watchOwnedObjects
This is a global variable that was introduced as a hotfix back when secrets used to be updated on each key. However, it introduces a race condition across threads. This logic is no longer needed.

jrcast:
1. Add finalizers to SAC and secrets
1. Remove `watchOwnedObjects` logic
1. Remove `mgr.GetFieldIndexer().IndexField()` installation

Closes #154
Closes #155

### DIFF
--- a/controllers/secretagentconfiguration_controller.go
+++ b/controllers/secretagentconfiguration_controller.go
@@ -39,7 +39,6 @@ import (
 	"github.com/ForgeRock/secret-agent/pkg/k8ssecrets"
 	corev1 "k8s.io/api/core/v1"
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // SecretAgentConfigurationReconciler reconciles a SecretAgentConfiguration object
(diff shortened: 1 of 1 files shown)


controllers/secretagentconfiguration_controller.go
```diff
@@ -78,10 +73,30 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
...
 		return ctrl.Result{}, err
 	}
-	// Stop watching for secret changes while the reconcile loop is running
-	watchOwnedObjects = false
-	defer func() { watchOwnedObjects = true }()
 	log.V(1).Info("** Reconcile loop start **")
+	saFinalizerName := "secret-agent.secrets.forgerock.io/" + instance.Namespace + "." + instance.Name
+
+	// check DeletionTimestamp to determine if the SAC is under deletion
+	if instance.ObjectMeta.DeletionTimestamp.IsZero() {
```
maxres-fr, line 80: I think this would read better if it were `if ! instance.ObjectMeta.DeletionTimestamp.IsZero()` then there would be no else required.

maxres-fr (approved):

wstrange (changes requested):
I dont understand the finalizer changes. Let's discuss before merging

PR #159, opened 2020-12-02
Refactor reconcile

maxres-fr:


wstrange (approved):
I'm liking it

controllers/secretagentconfiguration_controller.go
```diff
@@ -82,48 +83,48 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
...
-	//		log.Error(err, "error loading cloud credentials from secret provided",
-	//			"secret_name", instance.Spec.AppConfig.CredentialsSecretName,
-	//			"namespace", cloudCredNS)
-	//		return ctrl.Result{}, err
-	//	}
-
-	//}
+	//	secretMgr := secretmanager.NewSecretManager(rclient, instance.AppConfig)
+
+	if instance.Spec.AppConfig.SecretsManager != v1alpha1.SecretsManagerNone &&
```
wstrange, line 88: worth moving all this to a function?

controllers/secretagentconfiguration_controller.go
```diff
@@ -145,26 +146,28 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
 		// if the secret had data we shouldn't do anything
 		if len(secObject.Data) != 0 {
 			// TODO log warning secret exists
+			// this isn't an error
 			// secret exists, skip
 			errorFound = true
```
wstrange, line 151: rename to skipSecret or something like it?  the errorFound is confusing

pkg/generator/generator.go
```diff
@@ -53,205 +57,239 @@ type keyGenConfig struct {
...
-	for key := range queue {
-		log.V(1).Info("reconciling secret key", "secret_name", secretReq.Name,
-			"data_key", key.Name, "secret_type", string(key.Type))
-
-		keyGenerator, err := generator.NewKeyGenerator(secretReq.Name, instance.Namespace, key, &instance.Spec.AppConfig)
-		if err != nil {
-			log.V(0).Info("error routing secret key type",
-				"secret_name", secretReq.Name,
+	g.Log = g.Log.WithValues("secret_name", g.SecObject.Name, "namespace", g.Namespace)
+	requeue, enquerer := makeKeyQueue(g.KeysToGen)
```
wstrange, line 63: maybe just queue instead of enquerer?

pkg/generator/generator.go
```diff
@@ -53,205 +57,239 @@ type keyGenConfig struct {
...
 func newKeyGenerator(
-	key *v1alpha1.KeyConfig, genConfig *GenConfig) (*keyGenConfig, error) {
+	key *v1alpha1.KeyConfig,
+	genConfig *GenConfig) *keyGenConfig {
 
 	var keyInterface KeyMgr
 	switch key.Type {
 	case v1alpha1.KeyConfigTypeCA:
-		keyInterface = NewRootCA(key)
+		keyInterface, _ = NewRootCA(key)
```
wstrange, line 133: if we always ignore the error, should we change the return types?

wstrange: How could this fail ? If the NewXX can never fail, then perhaps err is not needed

pkg/generator/generator.go
```diff
@@ -53,205 +57,239 @@ type keyGenConfig struct {
...
-		if err != nil {
-			return err
-		}
-		if empty {
-			// load dependencies.
-			// missing secrets and keys are treated are errors and should stop generation
-			// self referencing deps can be resolved by re-running again later
-			reEnqueForKey, err := keyGenerator.ConfigureDependencies(ctx, rclient)
-			// check for NotFoundErr, for retry
+			log.V(1).Info("key started")
```
wstrange, line 79: key generation started

pkg/generator/generator.go
wstrange: hmm, no thats not what that variable is.... its a string?

wstrange (approved):
I find the logic hard to follow, but I assume it works. 

I think this should go through another round of refactoring once we get it working. The function and variables names, and return types could be reviewed to make it clearer what they do. 

Another thought:  Would it help to create a secret manager interface type, along with a NoOp implementation.  Could it simplify this to  just:  load/store from secret manager. The NoOp manager does nothing - but has the same interface. Just a thought

pkg/generator/generator.go
maxres-fr: There's a couple key types that could fail. e.g. we create a temporary directory for a keystore on init so all work is done in a single dir and then cleaned.

maxres-fr: I think you had brought that up before. We were catching them.

pkg/generator/generator.go
maxres-fr: `enquerer`, could be `enquererFn` it checks to make sure the self referenced key exits

controllers/secretagentconfiguration_controller.go
maxres-fr: https://github.com/ForgeRock/secret-agent/pull/159/commits/c6c73a70cb938aaaae4790069eb64e480762ebb4 I have it started in the package wasn't going to touch this section at the moment.

pkg/generator/generator.go
maxres-fr: I think you didn't get why creating an object would have a err.

maxres-fr (commented):
>  The function and variables names, and return types could be reviewed to make it clearer what they do.

I'm a bit blind eyed to it all at the moment. I don't disagree. Naming has been rather hard since things like "key" has so many meanings in this project.

pkg/generator/generator.go
maxres-fr: @wstrange I've updated this, take a look.

maxres-fr:
> I find the logic hard to follow, but I assume it works.
I made a few changes last night that should make things a bit simpler. Is there somewhere in particular you find it hard to follow?

maxres-fr:
> Another thought: Would it help to create a secret manager interface type, along with a NoOp implementation. Could it simplify this to just: load/store from secret manager. The NoOp manager does nothing - but has the same interface. Just a thought

I think that's great but should be incorporated during the rework of how secrets for the manager is loaded. What I have stubbed out is an and interface so a NoOp would cut some logic down!

controllers/secretagentconfiguration_controller.go
```diff
@@ -130,210 +134,74 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
...
-			"secret_namespace", instance.Namespace,
-			"secret_name", secretReq.Name)
+		}
+	}
 
-	} // end of secrets loop
 	// Only update the instance's status if there was a k8s operation.
 	// No k8s operation happen in the last reconcile loop. Update to "completed" if no updates and no errors occurred
-	if updatedK8sSecrets || rescheduleRetry || errorFound {
+	if rescheduleRetry || errorFound {
```
jrcast, line 195: You'll want to keep the `updatedK8sSecrets` variable. if updatedK8sSecrets == True and `rescheduleRetry = errorFound = false` That means you just wrote the last secret or set of secrets. That's when we update the status to `Completed`

pkg/generator/generator.go
jrcast: more like _key processing started_. At this point, we don't know if the key is going to be loaded form a secret manager or generated

jrcast (approved):

pkg/generator/generator.go
maxres-fr: only two initializers now throw errors

PR #161, opened 2020-12-08
Add a PodDisruptionBudget to the deployment

### LINKED ISSUE #160: Add a PodDisruptionBudget for the operator
Cluster autoscaling operations can terminate the `secret-agent` operator. Add a PDB to ensure there's a least one healthy `secret-agent` pod at all times.

jrcast:
1. Add a PodDisruptionBudget to the deployment

Closes #160

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,7 @@ release:
 	git push origin $(VERSION)
 	cd config/manager && kustomize edit set image controller=${IMG}
 	@mkdir -p dist/ && kustomize build config/default > dist/secret-agent.yaml
+	@kustomize build config/ha > dist/secret-agent-ha.yaml
 	curl https://proxy.golang.org/github.com/forgerock/secret-agent/@v/$(VERSION).info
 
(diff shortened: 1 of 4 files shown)


config/manager/pdb.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: controller-manager
+  namespace: system
```
wstrange, line 5: Is this the right namespace? 
Perhaps omit namespace and let kustomize set it?

Off topic, but it would be nice to move this into the fr-system namespace ;-)

wstrange (approved):

config/manager/pdb.yaml
jrcast: that ns is set to `system` just so that it matches all other manifests in the project. The namespace is set to `secret-agent-system` by the top level kustomization file.

wstrange (approved):

maxres-fr (approved):

PR #167, opened 2021-01-08
Refactor Secret Manager Client

### LINKED ISSUE #166: Secret Manager Refactor Client
The secret manager package in secret agent provides a couple methods to write and read data to all three providers secret managers. At the moment we pass secret manager access configuration all the way down from the controller down to a key manager which is unnecessary and cumbersome and hard to test.

Goal is to have an interface for the secret manager:

```
type SecretManager interface {
      EnsureSecret(ctx context.Context, secretName string, value []byte) error
      LoadSecret(ctx context.Context, secretName string) ([]byte, error)
  }
```

 
The interface should be created with a NewSecretManager function that takes an app config, adds the credentials etc, to a struct that is returned meeting the interface.

e.g.

```
  type secretManager struct {
      gcpClient   *secretmanager.Client
      awsClient   *awssecretsmanager.SecretsManager
      azureClient *keyvault.BaseClient
      config      *v1alpha1.AppConfig
  }

  func NewSecretManager(config) SecretManager {
// load creds
   return &secretManager{}
  }
```

 

1. Code from ~LOC L87-125 in controllers/secretagentconfiguration_controller.go should be lifted into the NewSecretManager func.
2. The controller should now call this new function then add the returned object to the GenKeysConfig and then secretManagerHasData and syncKeys will now call the load and ensure functions passing the interface instead of an appconfig object
3. update all ensure/load key manager interfaces use the new interface properly

lee-baines:
Create newSecretManager function to create a new cloud vendor specific SecretManager object.
Create new SecretManager interface to execute loading and adding of secrets and closing of clients.
Move all secret manager logic to the secretsmanager package.
Update generator packages to call the SecretManager interface.

Closes #166

### DIFF
--- a/controllers/secretagentconfiguration_controller.go
+++ b/controllers/secretagentconfiguration_controller.go
@@ -18,10 +18,6 @@ package controllers
 
 import (
 	"context"
-	"fmt"
-	"io/ioutil"
-	"os"
-	"path"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -36,6 +32,7 @@ import (
 
 	"github.com/ForgeRock/secret-agent/pkg/generator"
 	"github.com/ForgeRock/secret-agent/pkg/k8ssecrets"
+	"github.com/ForgeRock/secret-agent/pkg/secretsmanager"
 	corev1 "k8s.io/api/core/v1"
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
 )
@@ -67,7 +64,10 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
 	)
 
 	var instance v1alpha1.SecretAgentConfiguration
-	if err := reconciler.Get(ctx, req.NamespacedName, &instance); err != nil {
+	var sm secretsmanager.SecretManager
+
+	err := reconciler.Get(ctx, req.NamespacedName, &instance)
+	if err != nil {
 		if k8serror.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
@@ -79,51 +79,16 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
 	}
 	log.V(1).Info("** Reconcile loop start **")
 
-	// TODO remove this and move setup of secret manager to a container
-	// add conatiner to GenKeyConfig
+	cloudCredNS := reconciler.CloudSecretsNamespace
 
-	//	secretMgr := secretmanager.NewSecretManager(rclient, instance.AppConfig)
-
-	if instance.Spec.AppConfig.SecretsManager != v1alpha1.SecretsManagerNone &&
-		instance.Spec.AppConfig.CredentialsSecretName != "" {
-		var dir string
-		var cloudCredNS string
-
-		if len(reconciler.CloudSecretsNamespace) > 0 {
-			cloudCredNS = reconciler.CloudSecretsNamespace
-		} else {
-			cloudCredNS = instance.Namespace
-		}
-
-		// load credentials secret
-		secObject, err := k8ssecrets.LoadSecret(reconciler.Client,
-			instance.Spec.AppConfig.CredentialsSecretName, cloudCredNS)
-		if err != nil {
-			log.Error(err, "error loading cloud credentials secret from the Kubernetes API",
-				"secret_name", instance.Spec.AppConfig.CredentialsSecretName,
-				"cloud_secret_namespace", cloudCredNS)
-			return ctrl.Result{}, err
-		}
-		// Create temporary dir for gcp credentials if needed
-		if instance.Spec.AppConfig.SecretsManager == v1alpha1.SecretsManagerGCP {
-			dir, err = ioutil.TempDir("", "cloud_credentials-*")
-			if err != nil {
-				log.Error(err, "couldn't create a temporary credentials dir")
-				return ctrl.Result{}, err
-			}
-			// clean up after ourselves
-			defer os.RemoveAll(dir)
-		}
-
-		// load cloud credentials to envs and/or files
-		if err := manageCloudCredentials(instance.Spec.AppConfig.SecretsManager, secObject, dir); err != nil {
-			log.Error(err, "error loading cloud credentials from secret provided",
-				"secret_name", instance.Spec.AppConfig.CredentialsSecretName,
-				"cloud_secret_namespace", cloudCredNS)
-			return ctrl.Result{}, err
-		}
-
+	// Create new Secret Manager object
+	if ctx, sm, err = secretsmanager.NewSecretManager(ctx, &instance, cloudCredNS, reconciler.Client); err != nil {
+		return ctrl.Result{}, err
 	}
+
+	// Close client
+	defer sm.CloseClient()
+
 	// set the SAC status to inProgress only the first time around.
 	if instance.Status.State == "" {
 		if err := reconciler.updateStatus(ctx, &instance, true, false); err != nil {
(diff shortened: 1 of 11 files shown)


controllers/secretagentconfiguration_controller.go
```diff
@@ -79,51 +79,22 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
...
-	// TODO remove this and move setup of secret manager to a container
-	// add conatiner to GenKeyConfig
-
-	//	secretMgr := secretmanager.NewSecretManager(rclient, instance.AppConfig)
-
-	if instance.Spec.AppConfig.SecretsManager != v1alpha1.SecretsManagerNone &&
-		instance.Spec.AppConfig.CredentialsSecretName != "" {
-		var dir string
-		var cloudCredNS string
+	rClient := reconciler.Client
```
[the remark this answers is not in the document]
lee-baines: good point

controllers/secretagentconfiguration_controller.go
```diff
@@ -67,6 +64,9 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
 	)
 
 	var instance v1alpha1.SecretAgentConfiguration
+	var sm *secretsmanager.SecretManager
+	var err error
+
 	if err := reconciler.Get(ctx, req.NamespacedName, &instance); err != nil {
```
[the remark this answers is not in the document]
maxres-fr: what's wrong with?

```
err := reconciler.Get
if err != nil {
// thing
}
```

controllers/secretagentconfiguration_controller.go
```diff
@@ -67,6 +64,9 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
 	)
 
 	var instance v1alpha1.SecretAgentConfiguration
+	var sm *secretsmanager.SecretManager
```
maxres-fr, line 67: Since this is an interface, we don't need a pointer here.

controllers/secretagentconfiguration_controller.go
```diff
@@ -79,51 +79,22 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
...
-		// load cloud credentials to envs and/or files
-		if err := manageCloudCredentials(instance.Spec.AppConfig.SecretsManager, secObject, dir); err != nil {
-			log.Error(err, "error loading cloud credentials from secret provided",
-				"secret_name", instance.Spec.AppConfig.CredentialsSecretName,
-				"cloud_secret_namespace", cloudCredNS)
-			return ctrl.Result{}, err
+	// Close client after reconcile loop finishes
+	defer func() {
+		if err := (*sm).CloseClient(); err != nil {
+			log.Error(err, "Problem closing client")
```
maxres-fr, line 94: why not `defer sm.CoseClient()`

pkg/generator/certificate.go
```diff
@@ -156,32 +156,32 @@ func (kp *CertKeyPair) References() ([]string, []string) {
...
 
 // LoadSecretFromManager populates RootCA data from secret manager
-func (kp *CertKeyPair) LoadSecretFromManager(ctx context.Context, config *v1alpha1.AppConfig, namespace, secretName string) error {
+func (kp *CertKeyPair) LoadSecretFromManager(ctx context.Context, sm *secretsmanager.SecretManager, namespace, secretName string) error {
 	var err error
 	publicPemKeyFmt := fmt.Sprintf("%s_%s_%s.pem", namespace, secretName, kp.Name)
 	privatePemKeyFmt := fmt.Sprintf("%s_%s_%s-private.pem", namespace, secretName, kp.Name)
 
-	kp.Cert.CertPEM, err = secretsmanager.LoadSecret(ctx, config, publicPemKeyFmt)
+	kp.Cert.CertPEM, err = (*sm).LoadSecret(ctx, publicPemKeyFmt)
```
maxres-fr, line 164: why are these calls being casted down? also wouldn't need to de reference... with my earlier comment.

maxres-fr (changes requested):

controllers/secretagentconfiguration_controller.go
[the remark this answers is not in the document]
lee-baines: Its because of this line: 
if ctx, sm, err = secretsmanager.NewSecretManager(ctx, &instance, cloudCredNS, rClient); err != nil {

Because of the 3 return values, it needs to be declared before.

controllers/secretagentconfiguration_controller.go
```diff
@@ -79,51 +79,22 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
...
-			dir, err = ioutil.TempDir("", "cloud_credentials-*")
-			if err != nil {
-				log.Error(err, "couldn't create a temporary credentials dir")
-				return ctrl.Result{}, err
-			}
-			// clean up after ourselves
-			defer os.RemoveAll(dir)
-		}
+	// Create new Secret Manager object
+	if ctx, sm, err = secretsmanager.NewSecretManager(ctx, &instance, cloudCredNS, rClient); err != nil {
```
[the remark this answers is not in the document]
lee-baines: Updated

controllers/secretagentconfiguration_controller.go
lee-baines: Because this was the only way I can capture the error. You can only defer 1 thing

controllers/secretagentconfiguration_controller.go
lee-baines: Doesn't seem to work without it?

lee-baines: Just pushed a commit removing the pointer.  Look better?

pkg/generator/certificate.go
lee-baines: Updated along with previous commit

controllers/secretagentconfiguration_controller.go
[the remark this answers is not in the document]
maxres-fr: I see... that's happening because 70 is an inline if, and the `err` is scoped  to the if statement.  If you use the "long" way it will be defined and it doesn't have to be declared.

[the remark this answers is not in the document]
maxres-fr: https://play.golang.org/p/bnPC5s_Bxwq

controllers/secretagentconfiguration_controller.go
maxres-fr: can't you just log the error in the closeclient method? You're not doing anything here with it, right?

pkg/generator/certificate.go
maxres-fr: you still have the type casting `(sm).LoadSecret` that shouldn't be needed.

controllers/secretagentconfiguration_controller.go
lee-baines: Yes you're right. I'll change that.

pkg/generator/certificate.go
lee-baines: fixed

maxres-fr (approved):
Thanks @lee-baines! :partying_face:

maxres-fr:
@lee-baines note we do require signed commits and a rebase/squash before merging!

pkg/generator/generator.go
```diff
@@ -29,23 +30,24 @@ var (
 type KeyMgr interface {
 	References() ([]string, []string)
 	LoadReferenceData(data map[string][]byte) error
-	LoadSecretFromManager(context context.Context, config *v1alpha1.AppConfig, namespace, secretName string) error
-	EnsureSecretManager(context context.Context, config *v1alpha1.AppConfig, namespace, secretName string) error
+	LoadSecretFromManager(context context.Context, sm secretsmanager.SecretManager, namespace, secretName string) error
```
jrcast, line 33: SecretManager should be available in `GenConfig` you shouldn't need to pass it as an arg

jrcast: ignore. I don't think this is correct.

jrcast (approved):

jrcast:
@lee-baines friendly reminder to squash commits when you merge.

PR #168, opened 2021-01-21
Remove dependency on ENV VARs to store client credentials

lee-baines:
Change using ENV VARs to store Secret Manager client credentials.
Instead use secret file directly in GCP Secret Manager client code. Use secret objects directly in AWS/Azure Secret Manager client code.
Remove manageCloudCredentials function and migrate code into the new Secret Manager object functions.

Closes CLOUD-2795

### DIFF
--- a/pkg/secretsmanager/secretsmanager.go
+++ b/pkg/secretsmanager/secretsmanager.go
@@ -8,14 +8,15 @@ import (
 	"path"
 
 	"strings"
-	"time"
 
 	secretmanager "cloud.google.com/go/secretmanager/apiv1beta1"
 	"github.com/Azure/azure-sdk-for-go/profiles/latest/keyvault/keyvault"
 	azauth "github.com/Azure/azure-sdk-for-go/services/keyvault/auth"
 	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/prometheus/common/log"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -23,6 +24,7 @@ import (
 	awssecretsmanager "github.com/aws/aws-sdk-go/service/secretsmanager"
 
 	"github.com/pkg/errors"
+	"google.golang.org/api/option"
 	secretspb "google.golang.org/genproto/googleapis/cloud/secretmanager/v1beta1"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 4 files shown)


pkg/secretsmanager/secretsmanager.go
```diff
@@ -409,75 +517,25 @@ func (sm *secretManagerNone) LoadSecret(ctx context.Context, secretName string)
...
+}
 
-		}
-		if keyValue, ok := secObject.Data[string(v1alpha1.SecretsManagerAzureClientSecret)]; ok {
-			if err := os.Setenv("AZURE_CLIENT_SECRET", string(keyValue)); err != nil {
-				return err
-			}
-		}
+// GetCredentialData extracts the credential data from the data field
+func GetCredentialData(secObject *corev1.Secret) ([]byte, error) {
```
jrcast, line 533: You probably don't need to expose this function. 
```suggestion
func getCredentialData(secObject *corev1.Secret) ([]byte, error) {
```

jrcast (approved):
LGTM

pkg/secretsmanager/secretsmanager.go
```diff
@@ -73,54 +76,27 @@ type secretManagerNone struct {
...
-		println(dir)
-		if err != nil {
-			log.Error(err, "couldn't create a temporary credentials dir")
-			return ctx, nil, err
-		}
-		// clean up after ourselves
-		defer os.RemoveAll(dir)
 	case v1alpha1.SecretsManagerAWS:
-		ctx, sm = newAWS(ctx, config)
+		ctx, sm, err = newAWS(ctx, config, rClient, cloudCredNS)
```
maxres-fr, line 97: I realize this is from the last PR but what are we returning the context here?

pkg/secretsmanager/secretsmanager.go
```diff
@@ -129,43 +105,180 @@ func NewSecretManager(ctx context.Context, instance *v1alpha1.SecretAgentConfigu
...
+	// if credentials secret is provided
+	if config.CredentialsSecretName != "" {
+		// load credentials secret from Kubernetes secret
+		secObject, err := LoadCredentialsSecret(rClient, config, cloudCredNS)
+		if err != nil {
+			return nil, &secretManagerAWS{}, err
+		}
+
+		// extract access AWS_ACCESS_KEY_ID from Kubernetes secret
+		if keyValue, ok := secObject.Data[string(v1alpha1.SecretsManagerAwsAccessKeyID)]; ok {
```
maxres-fr, line 203: Is "not ok" handled?

maxres-fr (commented):
Looks good. Couple of minor comments, only the "not ok" is important to me.

pkg/secretsmanager/secretsmanager.go
lee-baines: Because the ctx gets used in the client creation then passed back down the controller.  Apparently it has to be passed through all functions. Can't be used in a struct.

pkg/secretsmanager/secretsmanager.go
lee-baines: ah yes, I missed that. I'll submit a commit handling those.

pkg/secretsmanager/secretsmanager.go
lee-baines: done

maxres-fr (approved):

PR #179, opened 2021-02-11
Fix spec.appConfig.createKubernetesObjects

### LINKED ISSUE #178: spec.appConfig.createKubernetesObjects does not work as expected
The SAC has the flag `spec.appConfig.createKubernetesObjects`. However, it is currently not used in the `secret-agent`

If this flag is set the `false`, the `secret-agent` will be able to pull keys directly from secret managers. Secret Agent would still generate keys and write to secret managers but not create them in Kubernetes

jrcast:
1. Fix `spec.appConfig.createKubernetesObjects`
1. Update `kubectl get` printed columns
1. Update status logic to separate requested and created secrets
1. Update readme

Closes #178


pkg/generator/generator.go
```diff
@@ -276,6 +288,16 @@ func (k *keyGenConfig) secretManagerHasData(ctx context.Context) (bool, error) {
 	return k.keyMgr.IsEmpty(), nil
 }
 
+// loadRefFromManager load ref from secret manager and determine if empty
+func (k *keyGenConfig) loadRefFromManager(ctx context.Context, refName, refKey string) ([]byte, error) {
```
maxres-fr, line 292: :fire:

dgoldssfo (approved):
LGTM

maxres-fr (approved):
:shipit: 

We need to update forgeops-cli to check this configuration. Should we update forgeops with new config? Maybe run prod profile like this?

jrcast:
> :shipit:
> 
> We need to update forgeops-cli to check this configuration. Should we update forgeops with new config? Maybe run prod profile like this?

filed https://github.com/ForgeRock/forgeops-cli/issues/55

wstrange (commented):
lgtm

PR #182, opened 2021-02-17
fix: handles conflcts with a retry

maxres-fr:


wstrange (commented):
Is the controller chasing its tail here? Updating status. which causes a conflict, etc.?

wstrange (approved):

controllers/secretagentconfiguration_controller.go
```diff
@@ -97,6 +97,9 @@ func (reconciler *SecretAgentConfigurationReconciler) Reconcile(req ctrl.Request
 	// set the SAC status to inProgress only the first time around.
 	if instance.Status.State == "" {
 		if err := reconciler.updateStatus(ctx, &instance, true, false); err != nil {
+			if k8serror.IsConflict(err) {
+				return ctrl.Result{Requeue: true}, nil
```
jrcast, line 101: can you add an info debug print here? something  indicating we couldn't update the status

jrcast (approved):
LGTM

maxres-fr:
> Is the controller chasing its tail here? Updating status. which causes a conflict, etc.?

I don't know what's causing this. We do update the objects status during the reconcile loop - however the update is supposed to update the instance we are using so I'm not sure why we'd encounter a conflict. I wonder if there's caching in the client code that's causing this. Maybe we should update kubebuilder and see if this resolves it. I did poke through operators and found that at least knative's update method handles conflicts magically.

PR #186, opened 2021-03-19
Add log entry when loading from secret manager

### LINKED ISSUE #185: Log obtaining secrets from a cloud secrets manager
I was replicating the use case where you use a cloud secrets manager (GCP secrets manager, in this case), and having trouble getting it to work. So I was using the log from the secret-agent-controller-manager pod to see if things were working.

When I did manage to get it to work, I could tell it was working, because the log messages were different, but it would have been really nice to get a log message (or a field in an existing log message) telling me the secret had been successfully obtained from the cloud secret manager.

So I'm requesting a small enhancement to the log file output - to log information when a secret is successfully obtained from GCP/AWS Secrets Manager or Azure Key Vault.

jrcast:
1. Add log entry when loading from secret manager
1. Change "preparing secret" to debug only

Closes #185

### DIFF
--- a/pkg/generator/generator.go
+++ b/pkg/generator/generator.go
@@ -120,6 +120,8 @@ func (g *GenConfig) GenKeys(ctx context.Context) error {
 					keysToWork = append(keysToWork, key)
 					continue
 				}
+			} else {
+				log.V(0).Info("loaded from secret manager")
 			}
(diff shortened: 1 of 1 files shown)


maxres-fr (approved):

dgoldssfo (approved):
Nice! Thanks!

PR #187, opened 2021-03-23
fix: Add resource to authorizer such that the audience on the token i…

maxres-fr:
…s set correctly

ref: cloud-2974


pkg/secretsmanager/secretsmanager.go
```diff
@@ -250,6 +251,7 @@ func newAzure(config *v1alpha1.AppConfig, rClient client.Client, cloudCredNS str
 		if err != nil {
 			return &secretManagerAzure{}, err
 		}
+		fmt.Printf("secret content %+v \n", secObject.Data)
```
wstrange, line 254: left over debug? Make we take this out for final image

wstrange (approved):

shankar-forgerock:
Great work. Thanks.

PR #189, opened 2021-04-07
fix: Certificate Key Usages

lee-baines:
Add ClientAuth to certificate template as required by SSL Key Pair.
Remove ServerAuth and ClientAuth usages from CA cert as is not relevant.
This is required for CLOUD-2920 - Implement Trust Manager Provider.

ref: CLOUD-3042


maxres-fr (approved):

dgoldssfo (approved):
Cool. Looks like no doc requirement.

PR #192, opened 2021-06-22
Update go pkgs

maxres-fr:
update to golang 1.16
use latest kubebuilder
use latest controller-gen
minor changes in make how go bin is used etc

WIP, builds and tests pass. Have to run on a cluster and make sure all is well still.


jrcast (approved):
LGTM

Dockerfile
```diff
@@ -61,9 +61,9 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -ldflags "-s -
 
 FROM openjdk:11-jre-slim as release
 
-RUN apt-get update && \                                                                               
-    DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y curl lsof net-tools && \ 
-    apt-get clean all                                                                                 
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y curl lsof net-tools && \
```
wstrange, line 65: do want to include apt-get ugprade?

wstrange (approved):

Dockerfile
jrcast: 🤷 we're grabbing the latest jre image up there so is not going to be that much behind. The only concern would be image size. upgrade tends to blow out the image size.

lee-baines (approved):

PR #194, opened 2021-07-01
add new kind workflow

maxres-fr:
Adds a Github action that creates a cluster via kind, installs the
operator then creates a CRD making sure that it reaches a status of
completed. Then makes sure all the secrets are created. It doesn't
validate content.


wstrange (approved):
nice

jrcast (approved):
LGTM

lee-baines (approved):

PR #196, opened 2021-07-13
feat: setting secretsManagerPrefix on a sac will no longer use a name…

maxres-fr:
…space in the key name

ref: cloud-3071

This is desired by PIT teams or anyone who wants to restore secrets from independent of namepsace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,9 +222,11 @@ All that is required is to provide a Kubernetes secret with the same name and sa
 There is a naming convention used by `secret-agent` to store and read secrets from the cloud secret managers. In general, the names follow the format:
 
 ```bash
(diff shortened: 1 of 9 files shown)


wstrange (approved):
So readers digest - this lets the user backup secrets in one namespace, restore to another? ie. namespace is no longer signficant??

Can you update the README and also comments on any samples sac files

jrcast (approved):

lee-baines (approved):
Thanks for this Max.  This helps simplify multi-cluster work further.

maxres-fr:
> So readers digest - this lets the user backup secrets in one namespace, restore to another? ie. namespace is no longer signficant??
Namespace is only insignificant if you set the SecretManagerPrefix. The problem was there was prefix + namepsace meaning it would only work if the secrets were going to the same namespace.

PR #199, opened 2021-07-14
fix: remove resource limits re-imposed by kubebuilder upgrade

maxres-fr:


jrcast (approved):

config/manager/manager.yaml
```diff
@@ -62,12 +62,5 @@ spec:
             port: 8081
           initialDelaySeconds: 5
           periodSeconds: 10
-        resources:
```
wstrange, line 65: do we want to remove, or just adjust higher?

Issue #203, opened 2021-08-12
Debug option seems to do nothing


mstrent:
I've been troubleshooting an issue trying to get on-prem ForgeOps / Secret Agent connected to Azure Key Vault. In the course of that I tried adding "--debug=true" and "--debug" to the command line options for the manager container. No change in verbosity was observed in the logs.

BTW the issue seemed to be caused by our Man in the Middle firewall which intercepts HTTPS sessions and presents certificates signed by our internal CA. And of course your Secret Agent image doesn't trust our CA, so it couldn't verify the certificate chain it saw when trying to hit the Azure Key Vault host. But the Secret Agent log messages were pretty ambiguous, and --debug didn't seem to help at all. This could be improved. :)

[...]

PR #206, opened 2021-08-18
fix: Flags parsed after logger opts initialized, leading to debug to …

maxres-fr:
…always be set to false

Left over upgrade bug kb3 upgrade. Parse flags before setting logger opts.

ref: #203


jrcast (approved):

wstrange (approved):
lgtm

PR #210, opened 2021-09-08
fix: Admission registration API version is out of date

maxres-fr:


controllers/webhook_cabundler.go
```diff
@@ -10,7 +10,7 @@ import (
 	"path/filepath"
 	"time"
 
-	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
```
jrcast, line 13: when was this added to k8s? In other words, what will be the minimum k8s version to support this secret-agent?

jrcast (approved):

lee-baines (approved):

controllers/webhook_cabundler.go
wstrange: I believe 1.16

wstrange (approved):

PR #221, opened 2022-02-21
fix: Create new secret when Azure Secret is disabled

pearj:
Support creating a new secret when an Azure Secret is disabled.

With Azure Key Vault secrets are now [soft deleted](https://docs.microsoft.com/en-us/azure/key-vault/general/soft-delete-overview), so deleting a secret doesn't have an effect until purged which defaults to 90 days. This means it isn't possible to get the secret agent to regenerate a secret, for example to add additional entries to a keystore.

Individual versions of secrets can however be disabled, and then creating a new secret simply creates a new version of that secret. So this change ignores the "SecretDisabled" inner service error, in addition to the existing "SecretNotFound" behaviour.

### DIFF
--- a/pkg/secretsmanager/secretsmanager.go
+++ b/pkg/secretsmanager/secretsmanager.go
@@ -509,8 +509,17 @@ func (sm *secretManagerAzure) LoadSecret(ctx context.Context, secretName string)
 
 	response, err := sm.client.GetSecret(ctx, fmt.Sprintf(azureVaultURLFmt, sm.azureVaultName), secretID, "")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pearj:
I should mention I'm very new to go, so I don't know if there is a better way to implement this.

wstrange:
Thanks for the PR Joel.  It looks OK to me - but I want @lee-baines to also give it a review

wstrange (approved):
LGTM

pearj:
Thanks Warren, do need to sign my commit? I see it says "The base branch requires all commits to be signed"

pearj:
I have amended the commit and signed it.

lee-baines (approved):
Looks good Joel, thanks.

PR #230, opened 2022-11-11
feat: POC: Integrate ForgeOps install into terraform modules.

paulbsch:
Get helm chart version from Chart.yaml.

ref: CLOUD-4105


lee-baines (approved):
Looks Good

PR #238, opened 2023-06-02
Update base openjdk image and remove curl

lee-baines:


RobinBanbury-FR (approved):
Changes look fine to me - curl is not used so seems safe to remove.

PR #267, opened 2024-05-02
Revert v1.2.2 changes that caused some breakage.

paulbsch:


lee-baines (approved):

lee-baines:
You can ignore the mend and snyk security checks as I'd expect those to fail

pearj:
What broke in v1.2.2? I'm use Azure as a backend, and I'd already updated some of our environments to v1.2.2, I think I had some issues, but I fixed them by deleting the namespace that secret-agent was running in and redeploying.

lee-baines:
Hi @pearj, there were some issues raised from customers saying the secrets defined in the secretagentconfiguration weren't automatically being backed up to Azure Vault. There was also some issues with AWS.  As v1.2.2 was quite a large change, we've had to revert it and reapply the patches separately.  Apologies for any disruption

lee-baines:
Yes would be sensible for now, then look out for an updated release soon

PR #272, opened 2024-05-28
ForgeOps-4394 secret manager v2

lee-baines:
Cherry-pick @dnitsch's earlier PR to update secret-agent to work with secret manager v2 but create env var for AWS region so the secretmanager client can read the configured value in the secretagentconfiguration


gandru (approved):
LGTM

pearj:
I just noticed this commit also commits the `kind` binary. Was that on purpose?

lee-baines:
Good spot @pearj , I'll get this removed

PR #276, opened 2024-07-04
fix: Update 2nd openjdk entry

lee-baines:
Update 2nd openjdk entry.

ref: FORGEOPS-5166


lee-baines:
Missed the 2nd openjdk entry update in the previous PR so this fixes that

gandru (approved):

PR #285, opened 2025-03-19
Fix high and medium security vulnerabilities.

paulbsch:
ref: FORGEOPS-5745


paulbsch:
Updated image built and pushed to: us-docker.pkg.dev/forgeops-public/images/secret-agent:v1.2.5

lee-baines (approved):

PR #288, opened 2025-04-16
feat(docker) + use correct version of jdk and build arm version of image

robinMcA:
service uses open jdk 22, 
on arm using jdk 23 image throws an error for one type of secrest. one would assume this would be the same on non arm platforms.


lee-baines:
Hi @robinMcA , apologies for missing this PR.  We had to update the OpenJDK version as part of the standard patching process to address vulnerabilities so we can't easily revert it.  It appears to be working fine on all of our clusters. Also the latest secret-agent image v1.2.7 is already multi-arch so no further changes required there. 

image

lee-baines:
In fact, we are now using openjdk 25 in v1.2.7 of secret-agent so please pull the latest version

PR #294, opened 2025-08-19
Use latest debian stable to resolve vulnerabilities.

paulbsch:
ref: FORGEOPS-6048


lee-baines:
Also add forgeops changelog update for new secrt-agent version

PR #295, opened 2025-11-06
fix: Update OpenJDK version as current version is not available

lee-baines:
Current version of OpenJDK 25 is not available hence is breaking PR builds.

ref: FORGEOPS-6198


damonconway (approved):
LGTM

paulbsch (approved):

PR #298, opened 2025-12-23
Fips bcfks support

rjaxin:
Enable secret-agent to create bouncy castle keystores


lee-baines (approved):
Changes look ok

lee-baines:
The integration test stage has failed when building the image.  There is a `/bin/sh: 1: /usr/local/go/bin/go: Exec format error` which implies its not building or using a multi-arch image

lee-baines:
Have created a Jira to investigate.  @paulbsch can you please take a look at this bug for Rob

lee-baines:
secret-agent is being deprecated so we will look to provide and alternative way to generate bouncycastle keystores

